PR #8, opened 2017-05-28
Add API for controlling Anki's GUI

glutanimate:
Hi there,

first of all thanks a lot for all the work you've put into AnkiConnect and all of your other Anki-related projects. They've been a great help to a great many people!

This commit extends the API with a couple of functions that allow invoking the AddCards and Browser window, respectively. I mostly wrote these for my [anki-cli-remote](https://github.com/glutanimate/anki-cli-remote) project, but I thought they might also find a general use in your add-on.

Though if these are outside the scope of the project I'll completely understand as well.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -421,7 +421,7 @@ class AnkiBridge:
         browser = aqt.dialogs.open("Browser", self.window())
         browser.activateWindow()
         if query:
-            query = unicode('"{}"'.format(query))
+            query = unicode('"{}"').format(query)
             browser.form.searchEdit.lineEdit().setText(query)
             browser.onSearch()
         return browser.model.cards
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -417,20 +417,19 @@ class AnkiBridge:
             return collection.decks.allNames()
 
 
-    def launchBrowser(self, query):
+    def guiBrowse(self, query):
         browser = aqt.dialogs.open("Browser", self.window())
         browser.activateWindow()
         if query:
-            query = u'"{}"'.format(query)
+            query = unicode('"{}"'.format(query))
             browser.form.searchEdit.lineEdit().setText(query)
             browser.onSearch()
         return browser.model.cards
 
 
-    def launchAddCards(self):
+    def guiAddCards(self):
         addcards = aqt.dialogs.open("AddCards", self.window())
         addcards.activateWindow()
-        return 0
 
 
 #
@@ -532,12 +531,12 @@ class AnkiConnect:
         return API_VERSION
 
 
-    def api_guiSearch(self, query):
-        return self.anki.launchBrowser(query)
+    def api_guiBrowse(self, query):
+        return self.anki.guiBrowse(query)
 
 
-    def api_guiAdd(self):
-        return self.anki.launchAddCards()
+    def api_guiAddCards(self):
+        return self.anki.guiAddCards()
 
 #
 #   Entry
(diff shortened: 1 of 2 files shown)


AnkiConnect.py
```diff
@@ -417,6 +417,22 @@ def deckNames(self):
...
+            query = u'"{}"'.format(query)
+            browser.form.searchEdit.lineEdit().setText(query)
+            browser.onSearch()
+        return browser.model.cards
+
+
+    def launchAddCards(self):
+        addcards = aqt.dialogs.open("AddCards", self.window())
+        addcards.activateWindow()
+        return 0
```
FooSoft: Prefer to return `None` instead of an arbitrary value of `0`. This can be done implicitly by just deleting `return 0` (don't have to add `return None`.

AnkiConnect.py
```diff
@@ -417,6 +417,22 @@ def deckNames(self):
             return collection.decks.allNames()
 
 
+    def launchBrowser(self, query):
+        browser = aqt.dialogs.open("Browser", self.window())
+        browser.activateWindow()
+        if query:
+            query = u'"{}"'.format(query)
```
FooSoft: The Anki alpha uses Python 3 which does not define `u` prefix for unicode strings (all strings are unicode by default), so this code will break. Use the `unicode` wrapper defined in the file instead:
```
query = unicode('"{}"'.format(query))
```

FooSoft:
I think that these changes can be useful so it makes sense to merge them in (especially the browser one). Please look over my comments about a couple of small changes that need to be made to the code. I would also appreciate if you added documentation to the `README.md` that explains the usage of these functions to other programmers who want to code against the AnkiConnect API.

glutanimate:
Thanks for the helpful comments! I'm glad to hear that you see these changes as useful.

I've just pushed a couple of commits that should hopefully address all of the remaining issues. Alongside fixing the problems you've pointed out I've also updated the function names to be more consistent with the rest of the API.

Please let me know if there are any other issues that remain.

AnkiConnect.py
```diff
@@ -417,20 +417,19 @@ def deckNames(self):
             return collection.decks.allNames()
 
 
-    def launchBrowser(self, query):
+    def guiBrowse(self, query):
         browser = aqt.dialogs.open("Browser", self.window())
         browser.activateWindow()
         if query:
-            query = u'"{}"'.format(query)
+            query = unicode('"{}"'.format(query))
```
FooSoft: You got the parens wrong on this one, it should be done as the code snipped I posted. Your code does a format operation on a normal string and then converts it to unicode. This will not work if `query` is unicode. What you need to do is take a unicode string and format **that**. This would be: `query = unicode('"{}"').format(query)`

README.md
```diff
@@ -292,6 +292,47 @@ rather than raw JSON. If you are writing raw requests be sure to send valid JSON
...
+    ```
+    {
+        action: 'guiAddCards',
+        params: {}
+    }
+    ```
+
+    *Sample response*:
+    ```
+    None
```
FooSoft: Python serializes `None` as `null` to JSON, so the sample response should be `null` instead.

FooSoft:
Just a couple of small knitpicks remaining, and then this is getting merged :+1:

glutanimate:
Sorry about the rookie mistakes :). Things should be fixed now.

FooSoft:
:+1:

FooSoft:
@glutanimate I've just uploaded the new version of the extension to AnkiWeb, so you can just tell your users to download the latest from there. I've also bumped the API version from 2 to 3.

glutanimate:
Awesome, I'll update my README. Thanks again for the guidance!

PR #9, opened 2017-05-28
Fix API params handler

vchagaev:
There was a problem with adding notes

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -445,9 +445,15 @@ class AnkiConnect:
 
     def handler(self, request):
         action = 'api_' + request.get('action', '')
+        params = request.get('params')
+        api_func = getattr(self, action)
+
         if hasattr(self, action):
             try:
-                return getattr(self, action)(**(request.get('params') or {}))
+                if not params:
+                    return api_func()
+
+                return api_func(params)
             except TypeError:
                 return None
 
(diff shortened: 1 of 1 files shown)


FooSoft:
Looks good, thank you for your contribution!

FooSoft:
Even though upon initial glance this looked fine, this actually immediately breaks Yomichan. All of the functions expect parameters with the same names as those in the `params` object. Your change just passes the objects around directly, breaking most `canAddNotes` and probably almost everything else.

I'm reverting this change for now.

vchagaev:
Oh, sorry!
I think in this case we should fix readme because it contains a wrong example for 'addNote'.

FooSoft:
Are you referring to the sample JSON blob for `addNote` or the sample JavaScript function for using the API? Both look OK to me, but it is possible I'm missing something.

vchagaev:
This doesn't work for me. It returns null.

`curl localhost:8765 -X POST -d '{"action": "addNote", "params": {"deckName": "English", "modelName": "two", "fields": {"Front": "TestFront", "Back": "TestBack"}}}'`

FooSoft:
Looks like the docs were wrong, sorry about that, I will update them now. The command below will work:
```
curl localhost:8765 -X POST -d '{"action": "addNote", "params": {"note": {"deckName": "English", "modelName": "two", "fields": {"Front": "TestFront", "Back": "TestBack"}}}}'
```

FooSoft:
Docs have been updated. Please open an issue if you are still running into problems :+1:

PR #11, opened 2017-06-08
Add API for retrieving and answering cards

CharlesHenry:
API endpoints added + updated readme

- getNextCard
- showQuestion
- showAnswer
- answerCard
- checkState

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -28,7 +28,7 @@ import socket
 # Constants
 #
 
-API_VERSION = 3
+API_VERSION = 4
 TICK_INTERVAL = 25
 URL_TIMEOUT = 10
 URL_UPGRADE = 'https://raw.githubusercontent.com/FooSoft/anki-connect/master/anki_connect.py'
@@ -434,6 +434,59 @@ class AnkiBridge:
         addCards.activateWindow()
 
 
+    def getNextCard(self):
+        if (self.window().reviewer.card is None) or (self.window().state != "review"):
+            self.window().moveToState('review')
+
+        card = self.window().reviewer.card
+        nextCard = {}
+        nextCard['id'] = card.id
+        nextCard['question'] = card._getQA()['q']
+        nextCard['answer'] = card._getQA()['a']
+        nextCard['answerButtons'] = self.window().reviewer._answerButtonList()
+        nextCard['_model'] = json.dumps(card.note(reload)._model)
+        nextCard['ord'] = card.ord
+        nextCard['fields'] = json.dumps(card.note(reload).fields)
+        nextCard['_fmap'] = card.note(reload)._fmap
+        nextCard['success'] = 'true'
+        return json.dumps(nextCard)
+
+
+    def showQuestion(self):
+        self.window().reviewer._showQuestion()
+        return json.dumps({'success': 'true'})
+
+
+    def showAnswer(self):
+        if self.window().reviewer.mw.state != "review":
+            return json.dumps({'success': 'false', 'message': 'Window state is not review.'})
+        else:
+            self.window().reviewer._showAnswer()
+            return json.dumps({'success': 'true'})
+
+
+    def answerCard(self, id, ease):
+        if self.window().reviewer.mw.state != "review":
+            return json.dumps({'success': 'false', 'message': 'Window state is not review.'})
+        elif self.window().reviewer.state != "answer":
+            return json.dumps({'success': 'false', 'message': 'Reviewer state is not answer.'})
+        elif self.window().reviewer.card.id != id:
+            return json.dumps({'success': 'false', 'message': 'Given card does not match.'})
+        elif self.window().col.sched.answerButtons(self.window().reviewer.card) < ease:
+            return json.dumps({'success': 'false', 'message': 'Invalid ease provided.'})
+        else:
+            self.window().reviewer._answerCard(ease)
+            return json.dumps({'success': 'true'})
+
+
+    def checkState(self):
+        stateInfo = {}
+        stateInfo['success'] = 'true'
+        stateInfo['window_state'] = self.window().state
+        stateInfo['reviewer_state'] = self.window().reviewer.state
+        return json.dumps(stateInfo)
+
+
 #
 # AnkiConnect
 #
@@ -539,6 +592,25 @@ class AnkiConnect:
 
     def api_guiAddCards(self):
         return self.anki.guiAddCards()
+
+    def api_getNextCard(self):
+        return self.anki.getNextCard()
+
+
+    def api_answerCard(self, id, ease):
+        return self.anki.answerCard(id, ease)
+
+
+    def api_checkState(self):
+        return self.anki.checkState()
+
+
+    def api_showQuestion(self):
+        return self.anki.showQuestion()
+
+
+    def api_showAnswer(self):
+        return self.anki.showAnswer()
 
 #
 #   Entry
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -448,7 +448,7 @@ class AnkiBridge:
         nextCard['ord'] = card.ord
         nextCard['fields'] = json.dumps(card.note(reload).fields)
         nextCard['_fmap'] = card.note(reload)._fmap
-        nextCard['success'] = 'true'
+        nextCard['success'] = True
         return json.dumps(nextCard)
 
 
@@ -457,34 +457,34 @@ class AnkiBridge:
             self.window().moveToState('review')
 
         self.window().reviewer._showQuestion()
-        return json.dumps({'success': 'true'})
+        return json.dumps({'success': True})
 
 
     def showAnswer(self):
         if self.window().reviewer.mw.state != "review":
-            return json.dumps({'success': 'false', 'message': 'Window state is not review.'})
+            return json.dumps({'success': False, 'message': 'Window state is not review.'})
         else:
             self.window().reviewer._showAnswer()
-            return json.dumps({'success': 'true'})
+            return json.dumps({'success': True})
 
 
     def answerCard(self, id, ease):
         if self.window().reviewer.mw.state != "review":
-            return json.dumps({'success': 'false', 'message': 'Window state is not review.'})
+            return json.dumps({'success': False, 'message': 'Window state is not review.'})
         elif self.window().reviewer.state != "answer":
-            return json.dumps({'success': 'false', 'message': 'Reviewer state is not answer.'})
+            return json.dumps({'success': False, 'message': 'Reviewer state is not answer.'})
         elif self.window().reviewer.card.id != id:
-            return json.dumps({'success': 'false', 'message': 'Given card does not match.'})
+            return json.dumps({'success': False, 'message': 'Given card does not match.'})
         elif self.window().col.sched.answerButtons(self.window().reviewer.card) < ease:
-            return json.dumps({'success': 'false', 'message': 'Invalid ease provided.'})
+            return json.dumps({'success': False, 'message': 'Invalid ease provided.'})
         else:
             self.window().reviewer._answerCard(ease)
-            return json.dumps({'success': 'true'})
+            return json.dumps({'success': True})
 
 
     def checkState(self):
         stateInfo = {}
-        stateInfo['success'] = 'true'
+        stateInfo['success'] = True
         stateInfo['window_state'] = self.window().state
         stateInfo['reviewer_state'] = self.window().reviewer.state
         return json.dumps(stateInfo)
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -596,6 +596,7 @@ class AnkiConnect:
     def api_guiAddCards(self):
         return self.anki.guiAddCards()
 
+
     def api_getNextCard(self):
         return self.anki.getNextCard()
 
@@ -604,16 +605,16 @@ class AnkiConnect:
         return self.anki.answerCard(id, ease)
 
 
-    def api_checkState(self):
-        return self.anki.checkState()
-
-
     def api_showQuestion(self):
         return self.anki.showQuestion()
 
 
     def api_showAnswer(self):
         return self.anki.showAnswer()
+
+
+    def api_checkState(self):
+        return self.anki.checkState()
 
 #
 #   Entry
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -434,7 +434,7 @@ class AnkiBridge:
         addCards.activateWindow()
 
 
-    def getNextCard(self):
+    def guiGetNextCard(self):
         if self.window().reviewer.card is None or self.window().state != 'review':
             self.window().moveToState('review')
 
@@ -452,7 +452,7 @@ class AnkiBridge:
         }
 
 
-    def showQuestion(self):
+    def guiShowQuestion(self):
         if self.window().reviewer.card is None or self.window().state != 'review':
             self.window().moveToState('review')
 
@@ -460,7 +460,7 @@ class AnkiBridge:
         return {'success': True}
 
 
-    def showAnswer(self):
+    def guiShowAnswer(self):
         if self.window().reviewer.mw.state != 'review':
             return {'success': False, 'message': 'Window state is not review.'}
         else:
@@ -468,7 +468,7 @@ class AnkiBridge:
             return {'success': True}
 
 
-    def answerCard(self, id, ease):
+    def guiAnswerCard(self, id, ease):
         if self.window().reviewer.mw.state != 'review':
             return {'success': False, 'message': 'Window state is not review.'}
         elif self.window().reviewer.state != 'answer':
@@ -597,20 +597,20 @@ class AnkiConnect:
         return self.anki.guiAddCards()
 
 
-    def api_getNextCard(self):
-        return self.anki.getNextCard()
+    def api_guiGetNextCard(self):
+        return self.anki.guiGetNextCard()
 
 
-    def api_answerCard(self, id, ease):
-        return self.anki.answerCard(id, ease)
+    def api_guiAnswerCard(self, id, ease):
+        return self.anki.guiAnswerCard(id, ease)
 
 
-    def api_showQuestion(self):
-        return self.anki.showQuestion()
+    def api_guiShowQuestion(self):
+        return self.anki.guiShowQuestion()
 
 
-    def api_showAnswer(self):
-        return self.anki.showAnswer()
+    def api_guiShowAnswer(self):
+        return self.anki.guiShowAnswer()
 
 
     def api_checkState(self):
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -435,59 +435,59 @@ class AnkiBridge:
 
 
     def getNextCard(self):
-        if (self.window().reviewer.card is None) or (self.window().state != "review"):
+        if self.window().reviewer.card is None or self.window().state != 'review':
             self.window().moveToState('review')
 
         card = self.window().reviewer.card
-        nextCard = {}
-        nextCard['id'] = card.id
-        nextCard['question'] = card._getQA()['q']
-        nextCard['answer'] = card._getQA()['a']
-        nextCard['answerButtons'] = self.window().reviewer._answerButtonList()
-        nextCard['model_name'] = card.note(reload)._model['name']
-        nextCard['ord'] = card.ord
-        nextCard['fields'] = json.dumps(card.note(reload).fields)
-        nextCard['_fmap'] = card.note(reload)._fmap
-        nextCard['success'] = True
-        return json.dumps(nextCard)
+        return {
+            'success': True,
+            'id': card.id,
+            'question': card._getQA()['q'],
+            'answer': card._getQA()['a'],
+            'answerButtons': self.window().reviewer._answerButtonList(),
+            'modelName': card.note(reload)._model['name'],
+            'ord': card.ord,
+            'fields': card.note(reload).fields,
+            'fieldMap': card.note(reload)._fmap
+        }
 
 
     def showQuestion(self):
-        if (self.window().reviewer.card is None) or (self.window().state != "review"):
+        if self.window().reviewer.card is None or self.window().state != 'review':
             self.window().moveToState('review')
 
         self.window().reviewer._showQuestion()
-        return json.dumps({'success': True})
+        return {'success': True}
 
 
     def showAnswer(self):
-        if self.window().reviewer.mw.state != "review":
-            return json.dumps({'success': False, 'message': 'Window state is not review.'})
+        if self.window().reviewer.mw.state != 'review':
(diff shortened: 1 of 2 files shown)


FooSoft:
Thanks for the code (and the updated docs!), looks very useful. I will merge this in after the issues I've mentioned in the code are resolved.

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+        nextCard['model_name'] = card.note(reload)._model['name']
+        nextCard['ord'] = card.ord
+        nextCard['fields'] = json.dumps(card.note(reload).fields)
+        nextCard['_fmap'] = card.note(reload)._fmap
+        nextCard['success'] = 'true'
+        return json.dumps(nextCard)
+
+
+    def showQuestion(self):
+        if (self.window().reviewer.card is None) or (self.window().state != "review"):
```
FooSoft: Same issue with braces & double quotes

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+            self.window().moveToState('review')
+
+        self.window().reviewer._showQuestion()
+        return json.dumps({'success': 'true'})
```
FooSoft: How about you just return `None` if there is no error, otherwise a string with the error description? By the way, you can just leave out the return statement. No return at end of function is the same as `return None`

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+
+
+    def showAnswer(self):
+        if self.window().reviewer.mw.state != "review":
+            return json.dumps({'success': 'false', 'message': 'Window state is not review.'})
```
FooSoft: Same thing here, just do `return 'Window state is not review.'` or something

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+        else:
+            self.window().reviewer._showAnswer()
+            return json.dumps({'success': 'true'})
+
+
+    def answerCard(self, id, ease):
+        if self.window().reviewer.mw.state != "review":
```
FooSoft: For all of these, get rid of `json.dumps` and just return error as string.

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+        elif self.window().reviewer.card.id != id:
+            return json.dumps({'success': 'false', 'message': 'Given card does not match.'})
+        elif self.window().col.sched.answerButtons(self.window().reviewer.card) < ease:
+            return json.dumps({'success': 'false', 'message': 'Invalid ease provided.'})
+        else:
+            self.window().reviewer._answerCard(ease)
+            return json.dumps({'success': 'true'})
+
+
+    def checkState(self):
```
FooSoft: Look at my comments for `getNextCard`, apply the same fixes here

AnkiConnect.py
```diff
@@ -540,6 +596,26 @@ def api_guiBrowse(self, query):
     def api_guiAddCards(self):
         return self.anki.guiAddCards()
 
+
+    def api_getNextCard(self):
+        return self.anki.getNextCard()
+
+
+    def api_answerCard(self, id, ease):
```
FooSoft: prefix all GUI related operations with `gui`, same as in `api_guiAddCards`

CharlesHenry:
Thanks for the speedy response! I'll sort these changes out and get back to you soon.

AnkiConnect.py
```diff
@@ -434,60 +434,60 @@ def guiAddCards(self):
...
-    def showQuestion(self):
-        if (self.window().reviewer.card is None) or (self.window().state != "review"):
+        return {
+            'success': True,
+            'id': card.id,
+            'question': card._getQA()['q'],
+            'answer': card._getQA()['a'],
+            'answerButtons': self.window().reviewer._answerButtonList(),
+            'modelName': card.note(reload)._model['name'],
+            'ord': card.ord,
```
FooSoft: Still not sure what `card.ord` is. If it's something useful, it should be renamed as such, otherwise it should not be exposed in the API.

AnkiConnect.py
```diff
@@ -434,60 +434,60 @@ def guiAddCards(self):
...
+        }
+
+
+    def guiShowQuestion(self):
+        if self.window().reviewer.card is None or self.window().state != 'review':
             self.window().moveToState('review')
 
         self.window().reviewer._showQuestion()
-        return json.dumps({'success': 'true'})
+        return {'success': True}
```
FooSoft: See my note about the `success` returns.

AnkiConnect.py
```diff
@@ -434,60 +434,60 @@ def guiAddCards(self):
...
     def checkState(self):
-        stateInfo = {}
-        stateInfo['success'] = 'true'
-        stateInfo['window_state'] = self.window().state
-        stateInfo['reviewer_state'] = self.window().reviewer.state
-        return json.dumps(stateInfo)
+        return {
+            'success': True,
+            'window_state': self.window().state,
+            'reviewer_state': self.window().reviewer.state
```
FooSoft: Camel case (`camelCase`) keys, no underscores please.

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+        elif self.window().reviewer.card.id != id:
+            return {'success': False, 'message': 'Given card does not match.'}
+        elif self.window().col.sched.answerButtons(self.window().reviewer.card) < ease:
+            return {'success': False, 'message': 'Invalid ease provided.'}
+        else:
+            self.window().reviewer._answerCard(ease)
+            return {'success': True}
+
+
+    def checkState(self):
```
FooSoft: Needs to be named `guiCheckState`.

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+        return {
+            'success': True,
+            'window_state': self.window().state,
```
FooSoft: Should be named `windowState` and `reviewerState`

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+            'fieldMap': card.note(reload)._fmap
+        }
+
+
+    def guiShowQuestion(self):
+        if self.window().reviewer.card is None or self.window().state != 'review':
+            self.window().moveToState('review')
+
+        self.window().reviewer._showQuestion()
+        return {'success': True}
```
FooSoft: This can never fail, no need for return value.

AnkiConnect.py
```diff
@@ -434,6 +434,62 @@ def guiAddCards(self):
...
+        if self.window().reviewer.mw.state != 'review':
+            return {'success': False, 'message': 'Window state is not review.'}
+        else:
+            self.window().reviewer._showAnswer()
+            return {'success': True}
+
+
+    def guiAnswerCard(self, id, ease):
+        if self.window().reviewer.mw.state != 'review':
+            return {'success': False, 'message': 'Window state is not review.'}
```
FooSoft: As I mentioned in the previous review, to keep things consistent with the rest of the API, we should not return success state with `success` and `message` variables. Just return `True` or `False` for actions or `None` when the caller expects a result.

CharlesHenry:
Hello, sorry for dragging this one out. Found it hard to find time to finish this off. There may be a couple things that need clarification on response format so please let me know if there's anything that needs changing and I'll try to sort it out asap. Thanks.

FooSoft:
This is pretty close, I'm just making a couple of small changes locally, but it should be ready to go in next couple of days :+1:

FooSoft:
Ended up changing quite a bit more things than I initially thought I would, but this is a much improved API, I think. The latest version of the extension has been uploaded to AnkiWeb.

CharlesHenry:
Hello, thanks a bunch for merging the changes! You've removed some features in your recent changes and have unfortunately broken some of the functionality. Here are some of the issues I have identified which I consider major issues:

- You can no longer start a review session from the API, you have to manually do this in the GUI in order to start fetching cards.
- There is no longer a way to extract the front and back of cards (for the basic model for example, the 'answer' in the current API output contains the question as well as the answer).
- Validations removed from guiShowAnswer mean that it can fail but return true if the window state is not 'review'
- guiAnswerCard does not validate the card that is being answered, this means that if an app (that is using the API) is ready to answer a card, and the user progresses with the GUI, the API will try to answer the previous card but accidentally answer a different card.

I'm not sure why you've removed the guiCheckState function or the labels from the buttons in the output. Without guiCheckState there is no way to identify the failure reason of guiAnswerCard (since there is no error message in the output anymore) and the button labels can be used for display on an app that is using the API.

I think that it might be worth reverting the functions back to how they were in the branch and working from there as they were in a working state and had been tested. I can do this and remove the 'success: True' fields from guiGetNextCard and guiCheckState if you would like. I am happy to convert the error messages to False to keep the API consistent but I believe that it has lost it's robustness and is no longer fully functional in its current state.

tomasgodoi:
@CharlesHenry 
Very useful what you are doing. I was planning on making something similar so I can review my Anki remotely, maybe using a Smartphone app. My plan is to use a Docker image (https://hub.docker.com/r/txgio/anki/) so I can run the Anki desktop version as a back-end for a web UI. (I know there is Ankidroid, but I use lots of addons which doesn't work with it and I don't like the synchronization at all because it takes a lot of time with my big deck)
I can help you guys to improve this API so it can be useful for my needs.

FooSoft:
@CharlesHenry, in designing APIs, I try to keep to the following four rules:
* Perform exactly one logical action for each request.
* Do not expose internals which can change in future versions.
* Do not provide features that cannot be used in all situations.
* Have a limited but general purpose set of APIs.

With that in mind, here are the replies to your comments:

> You can no longer start a review session from the API, you have to manually do this in the GUI in order to start fetching cards.

The API in the pull request did not provide a way to specify which deck should be reviewed and instead would just (I assume) open the default. If you can come up with an easy way of specifying which deck to start reviewing (such as `guiStartReview(deckName)`) then we should use that. I did not like that APIs would automatically put you into review mode as this violates the "do only one thing per request" principle. Starting a review is a discrete action and should be treated as such.

> There is no longer a way to extract the front and back of cards (for the basic model for example, the 'answer' in the current API output contains the question as well as the answer).
This is a good point, I will make it so that the question and answer responses provide the expected values.

> Validations removed from guiShowAnswer mean that it can fail but return true if the window state is not 'review'
Again, this is the "do only one thing" principle. If the caller of the API wants to show the answer to the card but the application is not in the review state, we should not just randomly put the user into review mode to try to fix an invalid assumption on the side of the API user. The calling application should be notified that the request could not be completed (you aren't showing the question, so what sense does it make to show the answer)?

> guiAnswerCard does not validate the card that is being answered, this means that if an app (that is using the API) is ready to answer a card, and the user progresses with the GUI, the API will try to answer the previous card but accidentally answer a different card.
Calling `guiCurrentCard` and verifying that you are still on the card you thought you were on should work fine, right? None of the other APIs take the cardId, so it does not make sense that this one would either. 

If you tell me more about what kind of application you are building for interfacing with Anki, perhaps I can help with API design.

FooSoft:
@CharlesHenry, also, regarding the buttons, why do you care about what Anki uses for the button labels? The only thing that matter are the ease values for using `guiAnswerCard`. There is no point in sending this data across the API constantly since you can just use your own constants.

CharlesHenry:
Hello,

Firstly I'd like to say I very much appreciate you going out of your way to respond to all of my concerns, I have some responses below. Please let me know what you think?

I agree with your first point, we should implement `guiStartReview(deckName)` at the moment the deck has to be selected on the deck browser screen in order to change deck, so getting the current card would start a review on the currently selected deck (which defaults to the last one that was reviewed).

> This is not correct. The answer only contains the question if that is how you configured your card template. The template is the only thing that even knows what is the "answer" and what is the "question".

I'm not sure what version of Anki you are using, but the latest version (2.0.44) is not like that by default. Please see the [Anki docs](https://apps.ankiweb.net/docs/manual.html#basic-templates). Here you can see the 'default back template' is 
```
{{FrontSide}}

<hr id=answer>

{{Back}}
```
For this reason, I originally included card fields  (to determine back and front fields regardless of template) and field order (to determine card direction). I included the question, answer and model fields as this is stuff that I would expect to be returned in a call that gets the current card.

> Validations removed from guiShowAnswer mean that it can fail but return true if the window state is not 'review'

This issue here is that the function returns true rather than returning an error as it did before in my implementation. This happens if you try calling guiShowAnswer when on the deck browser screen.

> Calling guiCurrentCard and verifying that you are still on the card you thought you were on should work fine, right? None of the other APIs take the cardId, so it does not make sense that this one would either.

This hardly sounds efficient to me as it requires you to call multiple endpoints each time you want to answer a card. Since you already know what card you are trying to answer, it should not be an issue to send it in.

The app I've built allows you to review Anki cards via notifications, I can work around changes to keep the API consistent but I require the fields and field order in order to display the values properly.


I'd like to propose changes to `guiCurrentCard` to add more to the output (as it is in my branch), a bugfix to `guiShowAnswer` to return an error when not in 'review' state and implementing `guiStartReview(deckName)` so that we can start reviews from the API.

Please let me know your thoughts.

FooSoft:
> I'm not sure what version of Anki you are using, but the latest version (2.0.44) is not like that by default. Please see the Anki docs. Here you can see the 'default back template' is <snip>
For this reason, I originally included card fields (to determine back and front fields regardless of template) and field order (to determine card direction). I included the question, answer and model fields as this is stuff that I would expect to be returned in a call that gets the current card.

Yeah, I realize that that is the default template. My point is that people can modify the field names and template to be whatever they want them to be. You can even make a card that has a field called "Back" for the question and "Front" for the answer -- there is nothing stopping that. We cannot rely on the "Back" and "Front" fields existing at all. If you are aware of a way of getting the "question" and "answer" sides of cards that does not rely on default fields and templates, that would certainly be something that would make sense to add (perhaps I am missing some additional information).

Regarding the model fields, you can get that information with the existing `modelFieldNames` API; I don't believe that duplicating information for the sake of convenience in specific usage scenarios is not good design. Valid ease answer values are useful for the current card (and vary depending on scheduling), so that is why they are being returned. Labels are not useful because you can always derive them yourself with the function below:
```
function getLabelForEase(ease) {
    return {1: 'Again', 2: 'Good', 3: 'Easy', 4: 'Very easy'}[ease] || 'invalid';
}
```

>This hardly sounds efficient to me as it requires you to call multiple endpoints each time you want to answer a card. Since you already know what card you are trying to answer, it should not be an issue to send it in.

Playing devil's advocate here, wouldn't `guiShowAnswer` and `guiShowQuestion` have the same problem though? How do you know that you are showing answer (or question) for the correct card? I am not against adding an *optional* `cardId` field that can be set to `null` if you do not care about making sure that you are working on the card that you think you are. One of the things that I wanted to try to do with this API is to add gamepad support where an external process can get gamepad input and use it to answer cards. Since in my case, the user is interacting directly with Anki, why should I even have to get `guiCurrentCard` just so that I can get the `cardId` in order to call `guiAnswerCard`? You see what I'm saying? 

> I'd like to propose changes to guiCurrentCard to add more to the output (as it is in my branch), a bugfix to guiShowAnswer to return an error when not in 'review' state and implementing guiStartReview(deckName) so that we can start reviews from the API.

In `master`, `guiShowAnswer` (as well as `guiShowQuestion`) already returns `False` when not in review mode, so that should already be sufficient for your needs. Regarding the output of `guiCurrentCard`, I am definitely for adding additional information, but that data has to be there to solve a specific problem. As I mentioned before, if you can tell me how to get *only* the answer and question text strings **regardless** of field and template settings, I will add any data required. I just don't want to dump a blob of JSON on the caller and say "here is a whole bunch of serialized Anki state, go figure it out".

FooSoft:
@CharlesHenry, oh and yes, we should definitely add a `guiStartReview(deckName)` function (maybe it should actually be called `guiReview` now that I think about it, since there will not be a corresponding `guiStopReview`).

CharlesHenry:
Hello,

Yup, I agree with you, just a couple things though.

> In master, guiShowAnswer (as well as guiShowQuestion) already returns False when not in review mode

Now that you say this, I can see that the 'validation' issue that I mentioned is with `guiReviewActive`.
If you start a review, and then go back to the deck browser, `guiReviewActive` will still return true, as a `self.reviewer().card` is no longer null. Because of this, `guiCurrentCard` will return that card, assuming that the review state is active. The side effect of this is the issue I've raised with validation. The fix would be to add `and self.window().state == 'review'` to the check.

Regarding the `guiCurrentCard `output, I understand that models can have any number of fields on the front and back, for example, Japanese models may have the Expression, Reading and Meaning rather than the basic Front / Back type fields. I also have anticipated that model names could be anything that the user desires and does not always correlate to the actual template. That being said, if you want to implement an app that needs to display fields, without the template formatting and html from Anki, then the only possible way is to expose the fields and the field order, the field map will then allow you to work out the order in which to display the fields based on the field order value. This can be configured for Basic and Japanese based on number of fields, which I believe with cater for most users without needing configuration of some type. The only improvement I can suggest is stripping out unneeded fields from the field map output (removing fields such as 'font' ,'rtl', 'size' and 'sticky'). Do you have suggestions on how to display the fields separately (forward and backward) in a more efficient way?

FooSoft:
>Now that you say this, I can see that the 'validation' issue that I mentioned is with guiReviewActive.
If you start a review, and then go back to the deck browser, guiReviewActive will still return true, as a self.reviewer().card is no longer null. Because of this, guiCurrentCard will return that card, assuming that the review state is active. The side effect of this is the issue I've raised with validation. The fix would be to add and self.window().state == 'review' to the check.

Good catch, I will make the required fix.

>Regarding the guiCurrentCard output, I understand that models can have any number of fields on the front and back, for example, Japanese models may have the Expression, Reading and Meaning rather than the basic Front / Back type fields. I also have anticipated that model names could be anything that the user desires and does not always correlate to the actual template. That being said, if you want to implement an app that needs to display fields, without the template formatting and html from Anki, then the only possible way is to expose the fields and the field order, the field map will then allow you to work out the order in which to display the fields based on the field order value. This can be configured for Basic and Japanese based on number of fields, which I believe with cater for most users without needing configuration of some type. The only improvement I can suggest is stripping out unneeded fields from the field map output (removing fields such as 'font' ,'rtl', 'size' and 'sticky'). Do you have suggestions on how to display the fields separately (forward and backward) in a more efficient way?

So I'm looking through some object dumps:

```
>>> pp(aqt.mw.reviewer.card.model())
{u'css': u'.card {\n font-family: arial;\n font-size: 20px;\n text-align: center;\n color: black;\n background-color: white;\n}\n',
 u'did': 1,
 u'flds': [{u'font': u'Arial',
            u'media': [],
            u'name': u'Expression',
            u'ord': 0,
            u'rtl': False,
            u'size': 20,
            u'sticky': False},
           {u'font': u'Arial',
            u'media': [],
            u'name': u'Reading',
            u'ord': 1,
            u'rtl': False,
            u'size': 20,
            u'sticky': False},
           {'font': u'Arial',
            'media': [],
            'name': u'Meaning',
            'ord': 2,
            'rtl': False,
            'size': 20,
            'sticky': False}],
 u'id': u'1496420295324',
 u'latexPost': u'\\end{document}',
 u'latexPre': u'\\documentclass[12pt]{article}\n\\special{papersize=3in,5in}\n\\usepackage[utf8]{inputenc}\n\\usepackage{amssymb,amsmath}\n\\pagestyle{empty}\n\\setlength{\\parindent}{0in}\n\\begin{document}\n',
 u'mod': 1499040025,
 u'name': u'Basic',
 u'req': [(0, 'all', [0])],
 u'sortf': 0,
 u'tags': [],
 u'tmpls': [{u'afmt': u'{{FrontSide}}\n\n<hr id=answer>\n\n<p>{{Meaning}}</p>\n\n<p>{{Reading}}</p>',
             u'bafmt': u'',
             u'bqfmt': u'',
             u'did': None,
             u'name': u'Card 1',
             u'ord': 0,
             u'qfmt': u'{{Expression}}'}],
 u'type': 0,
 u'usn': -1,
 u'vers': []}
```

```
>>> pp(aqt.mw.reviewer.card.note().fields)
[u'note_expression', u'note_reading', u'note_meaning']
```

```
>>> pp(aqt.mw.reviewer.card.note()._fmap)
{u'Expression': (0,
                 {u'font': u'Arial',
                  u'media': [],
                  u'name': u'Expression',
                  u'ord': 0,
                  u'rtl': False,
                  u'size': 20,
                  u'sticky': False}),
 u'Meaning': (2,
              {'font': u'Arial',
               'media': [],
               'name': u'Meaning',
               'ord': 2,
               'rtl': False,
               'size': 20,
               'sticky': False}),
 u'Reading': (1,
              {u'font': u'Arial',
               u'media': [],
               u'name': u'Reading',
               u'ord': 1,
               u'rtl': False,
               u'size': 20,
               u'sticky': False})}
```

The `ord` value is pretty much useless for what you want to do. It is basically the order of the fields in the model, and not in the card (as you can see from the objects). Even if ordering was the same as on the card, it would still not help you determine where the question ends and the answer begins.

How about I give you back a dictionary (stored in `fields` returned from `guiCurrentCard` that looks like this (for the objects above):
```
{{"Expression": "note_expression", "Meaning": "note_meaning", "Reading": "note_reading"}}
```

Since this is a dictionary, you would not have ordering information (which is useless anyway), but you would have the fields and values. In your application, you then could check for specific field names to detect common models.

CharlesHenry:
Hello,

Your solution does not allow me to work out if a card should be displayed Front/Back or Back/Front.

To identify the order of the fields (which can changed if the card is reviewed in both directions), you can use card.ord, to find the first field based on the fieldmap ord. Eg when card.ord = 0 then the the index in the fieldsmap array with ord = 0 is the first one, if card.ord = 1 then index in fieldmap where ord = 1 is the first one to display (fieldmap ord is not always the same as the index or index+1 so we need the actual value in the array). This works for basic models, the logic is a little more complicated when it comes to Japanese type models (you have to base off the order and field name).

This is the reason why I have `fieldOrder`, `fields`, and `fieldMap `in my original implementation.

Is this enough for you to consider putting these fields back into the output?

FooSoft:
Your explanation makes sense. 

So it sounds like getting `fieldOrder` and `fields` that look like `{{"Expression": {"value": "note_expression", "order": 0}, "Meaning": {"value": "note_meaning", "order": 1}, "Reading": {"value": "note_reading", "order": 2}}`

Where `fieldOrder` comes from `aqt.mw.reviewer.card` and the `order` in `fields` comes from `aqt.mw.reviewer.card.note()._fmap`?

CharlesHenry:
I guess it could be implemented that way. Based on the `fieldOrder` value, you can identify which value should be displayed first in `fields` (for basic models anyway).

FooSoft:
API and docs updated, extension uploaded to AnkiWeb.

CharlesHenry:
This is awesome, thanks a bunch!

FooSoft:
:+1:

PR #12, opened 2017-07-02
Include actions to open the Deck Overview for a given deck name and go to the Deck Browser

tomasgodoi:
Those actions would be useful if one wants to control user review. With it, you can go to a specific deck so you can review it (with the actions included by @CharlesHenry on his fork) and you can also go back to the Deck Browser screen, so you can stop the review process.

I don't know to which extent the author wants to increase the number of actions available. I'll need to implement full automation of the review process. If the author wants to keep his codebase simple, just reject this pull request and I can work my API needs on my fork.

By the way, it would be nice if we reestructure the code so each action can reside on its own file. I think that way would be easier to maintain in the case the API grows.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -507,6 +507,19 @@ class AnkiBridge:
         reviewer._answerCard(ease)
         return True
 
+    def guiDeckOverview(self, name):
+        collection = self.collection()
+        if collection is not None:
+            deck = collection.decks.byName(name)
+            if (deck is not None):
+                collection.decks.select(deck['id'])
+                self.window().onOverview()
+                return True
+        return False
+
+    def guiDeckBrowser(self):
+        self.window().moveToState("deckBrowser")
+        return True
 
 #
 # AnkiConnect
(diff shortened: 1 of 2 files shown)


AnkiConnect.py
```diff
@@ -507,6 +507,19 @@ def guiAnswerCard(self, ease):
         reviewer._answerCard(ease)
         return True
 
+    def guiDeckOverview(self, name):
+        collection = self.collection()
+        if collection is not None:
+            deck = collection.decks.byName(name)
+            if (deck is not None):
```
FooSoft: Don't need parenthesis around condition.

README.md
```diff
@@ -390,6 +390,40 @@ Below is a list of currently supported actions. Requests with invalid actions or
     ```
     true
     ```
+*   **guiDeckOverview**
+
+    Opens the Deck Overview screen for the deck with the given name; returns `true` if succeeded or `False` otherwise.
```
FooSoft: Should be lower-case `false` instead of `False`

README.md
```diff
@@ -390,6 +390,40 @@ Below is a list of currently supported actions. Requests with invalid actions or
...
+    ```
+
+    *Sample response*:
+    ```
+    true
+    ```
+
+*   **guiDeckBrowser**
+
+    Opens the Deck Browser screen; returns `true` if succeeded or `null` otherwise.
```
FooSoft: Please make the API return `true` and `false` instead of `true` and `null` for consistency.

tomasgodoi:
I made the requested fixes. I actually changed the guiDeckBrowser to return nothing like the guiAddCards action, because if I'm not mistaken it will never fail.

FooSoft:
Cool, thanks for the code :+1:

PR #15, opened 2017-07-21
Add fast non-GUI browse function

techdavid:
I'm writing a Memrise-style web frontend to Anki and I've found your amazing library, but the GUI functions are too slow when running them in quick succession (and they also try to bring Anki to the foreground). For those reasons I've started work on fast equivalents which don't involve the GUI.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -456,6 +456,13 @@ class AnkiBridge:
         return browser.model.cards
 
 
+    def browse(self, query=None):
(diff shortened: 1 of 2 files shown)


FooSoft:
This looks really good -- could you please also update the `README.md`?

techdavid:
No problem, I've just updated it and I'll try to remember next time I add a new function.

FooSoft:
Looks good, thanks!

PR #16, opened 2017-07-27
Add multi function

techdavid:
Another "performance" function - in the web client I'm writing, I've been using the browse function three times in a row to find the number of total, learned and ready-for-review cards for each deck. The multi action allows me to do all three actions in one ajax request, making it much faster.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -428,6 +428,13 @@ class AnkiBridge:
                 return [field['name'] for field in model['flds']]
 
 
+    def multi(self, actions):
+        response = []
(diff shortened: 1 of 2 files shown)


FooSoft:
This makes a lot of sense -- nice!

techdavid:
@FooSoft Thanks for the merge! By the way, I just put the project I'm doing on my GitHub -- if you want to have a look: https://github.com/techdavid/AnkiNouveau

So far it just lists decks and you can't learn anything, but hopefully it will be more functional soon

PR #17, opened 2017-07-30
Fix bug where guiDeckReview always returns null

techdavid:
The README says that guiDeckReview returns true or false, but the webapi function doesn't return anything and the README's sample response shows `null`.


FooSoft:
Looks good

PR #18, opened 2017-08-03
Add tagging and suspend functions

techdavid:
addTags, removeTags, suspend, unsuspend

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -375,6 +375,20 @@ class AnkiBridge:
             return note
 
 
+    def addTags(self, query, tags, add=True):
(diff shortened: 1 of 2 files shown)


FooSoft:
I like like the idea behind these functions but I think that they need to take IDs instead of queries in order to keep query related behavior isolated to one API. From your point of view it will require invoking `browse` before making the invocation, but I think it makes more sense in general as an interface.

techdavid:
Updated - I've also created a findNotes function, and renamed "browse" to "findCards" to match this naming convention.

FooSoft:
Looks good, thanks for making the change 👍

PR #19, opened 2017-08-05
Call startEditing before / stopEditing after tagging

techdavid:
Didn't realize I had to do this initially - usually there's no problem, but if a card you're trying to edit is selected in the browser it can cause problems. Suspending is okay though


FooSoft:
Looks good

Issue #20, opened 2017-08-05
Calling suspend resets the progress of part-learned cards


techdavid:
Just posting this in case you can figure out the problem - I've been trying to make it work but haven't been successful.

When suspend is called it resets the progress of any part-learned cards (= new cards that have been marked good 10min, but not good 1d yet) in any deck, regardless which cards have been suspended. The unsuspend function doesn't cause this.

From looking at [the relevant file][1] in the Anki source code (line 1265) this means it's either caused by `remFromDyn` line 981 or `removeLrn` line 682 (probably the latter). The problem also occurs when calling the following in the dev console - but *does not* occur when suspending from the browser:

    aqt.mw.col.sched.suspendCards([])  # array can be empty, or have any CIDs - problem still occurs

When cards are suspended from the browser, [the onSuspend function][2] (line 1250) is run, which in turn runs [saveNow][3] (line 300 + 390 for saveTags). However, the editor object isn't accessible from add-ons unless you create the editor from within the add-on, and I get the feeling this isn't what's causing the problem.

I've tried wrapping the function with startEditing() / stopEditing() but this doesn't work. Any ideas what could be the problem?

[1]: https://github.com/dae/anki/blob/master/anki/sched.py#L1265
[2]: https://github.com/dae/anki/blob/master/aqt/browser.py#L1250
[3]: https://github.com/dae/anki/blob/master/aqt/editor.py#L300

FooSoft:
So it looks like `saveNow` takes a callback that is executed when the save completes. This invokes the following functions:

```

    def onSuspend(self):
        self.editor.saveNow(self._onSuspend)

    def _onSuspend(self):
        sus = not self.isSuspended()
        c = self.selectedCards()
        if sus:
            self.col.sched.suspendCards(c)
        else:
            self.col.sched.unsuspendCards(c)
        self.model.reset()
        self.mw.requireReset()
```

I see that it is calling `reset` on the model, which is not happening in the `startEditing` and `stopEditing` methods in AnkiConnect... Maybe it should?

techdavid:
With the last bug I found (i.e. the one fixed in #19) I saw a call to model.reset as well when tags were added from the browser, and was wondering if this was the problem - I actually emailed Damien about it since I wasn't sure what to do, but he said it was unrelated. (just forwarded you the email chain, sorry for the messed up formatting)

[...]

PR #21, opened 2017-08-06
Fix issue #20; add isSuspended function; update code style

techdavid:
I realized the problem only occurs when all the cards given are already suspended - in my webapp, I'm automatically suspending certain cards by state, and because I'm reloading frequently to test it it means that the suspend function is called again with the same cards.

This is why I didn't realize that normally it works correctly, and also why it isn't a problem in the card browser - you can only toggle the suspended state, not suspend cards which are already suspended.

I also put startEditing / stopEditing calls around the suspend calls, to make sure the deck overview numbers are updated (although these aren't needed functionally). Additionally I changed `aqt.mw.col` to `self.collection()` in the new functions I wrote to match the style of the others.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -377,15 +377,36 @@ class AnkiBridge:
 
     def addTags(self, notes, tags, add=True):
         self.startEditing()
-        aqt.mw.col.tags.bulkAdd(notes, tags, add)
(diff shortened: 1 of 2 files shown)


FooSoft:
Sweet, I'm glad you figured it out :+1:

PR #24, opened 2017-08-12
Add areDue function and change isSuspended to areSuspended

techdavid:
(and some bugfixes / style changes to match existing code)

`areDue` returns an array for whether each of the given cards is due, and `isSuspended` has been changed to `areSuspended` to make the functions more consistent.


FooSoft:
Good work :+1:

PR #25, opened 2017-08-13
Add getDecks and deleteDecks functions

techdavid:
`getDecks` returns an object indicating which decks the given cards belong to. `deleteDecks` should be self-explanatory


FooSoft:
:+1:

PR #26, opened 2017-08-14
Add lock functions

techdavid:
These allow applications using AnkiConnect to detect if another instance is already running, to prevent errors occurring.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -25,6 +25,7 @@ import select
 import socket
 import sys
 from time import time
+from threading import Timer
 
 
 #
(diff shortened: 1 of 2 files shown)


AnkiConnect.py
```diff
@@ -685,6 +686,26 @@ def guiDeckReview(self, name):
             return False
 
 
+    def setLock(self, timeout=0):
+        if hasattr(self, 'lockTimer'):
```
FooSoft: Can you just make `lockTimer` be always be present on the object and be initialized to `None` instead? I

FooSoft:
I have some mixed thoughts about this API because the locks are not enforced even though that is what they (almost by definition) are supposed to do. If the idea is to prevent well-behaving applications from issuing multiple conflicting requests, shouldn't the applications just make sure that they are not making calls from multiple instances? Also the idea of using a timeout seems kind of arbitrary...

techdavid:
> shouldn't the applications just make sure that they are not making calls from multiple instances?

The idea was that local HTML5 webapps can't communicate between different instances of themselves without some kind of server (and can't set local storage on the `file:` protocol).

However, I've now realised that because I'm using Tampermonkey to access AnkiConnect, I can use its storage to set a lock instead. So this can be closed if you like, as yes it's a bit pointless.

> Also the idea of using a timeout seems kind of arbitrary...

Continuing from the first idea, when these webapps are closed with the browser close tab button there is no way to send a final query to remove the lock, so a solution is to lock the deck repeatedly for short amounts of time.

PR #28, opened 2017-08-14
Add supportedActions action

### LINKED ISSUE #27: Propose standard for handling releases / managing API version
With more and more changes going into this plugin, I think we need to devise a way to handle releases so that we can ensure that the functionality we require is in a certain version, without incrementing the API version every change.

The scenario here is that my app checks that API version 4 is running, but since changes are going in frequently there is a possibility that the user may or may not have the specific change I need in their version, even though they have version 4.

Once solution is to increment the API version on demand, when a dependant app is ready for a code release, so that we can be sure that the required functionality is present?

techdavid:
Fixes #27 (also fixed typo in the README)

Please edit the README entry for this action if you feel my usage recommendation is wrong. Also, maybe now is time for an API version increment?

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -850,6 +850,21 @@ class AnkiConnect:
 
 
     @webApi
+    def supportedActions(self):
+        actions = {}
+        for action in dir(AnkiConnect):
+            handler = getattr(AnkiConnect, action)
+
+            if hasattr(handler, 'webApi'):
+                if (hasattr(handler, 'revision')):
+                    actions[action] = handler.revision
+                else:
+                    actions[action] = 1
+
+        return actions
+
+
+    @webApi
     def findNotes(self, query=None):
         return self.anki.findNotes(query)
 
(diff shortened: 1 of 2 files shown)


techdavid:
Example of how to indicate a new revision:

    @webApi
    def guiDeckReview(self, name):
        return self.anki.guiDeckReview(name)
    guiDeckReview.version = 2

FooSoft:
The problem of API versioning in AnkiConnect is something I have been thinking about for a while. I think when I was designing the extension I made the assumption that polling for API version is a good idea. The problem is that naively written applications will not ask for the version number (or only ask for it once). There is also the possible (although unlikely case) where the version check passes fine, but the user upgrades AnkiConnect afterwards... There is also the problem of handling function renaming.

What I propose is adding a `version` field to every request that the client sends. The combination of the `version` and `action` fields can be used to find the proper API regardless of what version the client software expects. If the `version` field is not provided, it will automatically default to the API version before this change was introduced.

I think the `webApi` decorator can be modified to add a version range as well as a list of API aliases (in case a method were to be renamed in the future. What do you think about this plan (I would be happy to hook it up myself)?

techdavid:
Sounds good to me 👍 I agree it would be best for you to implement it so it works how you expect.
cc @CharlesHenry

PR #29, opened 2017-08-14
Fix issue with AnkiConnect self update

CharlesHenry:
Self update was failing due to the download URL being wrong and the error message box not working properly.


FooSoft:
Wow, I could have sworn that this worked before... Thanks for the fix.

PR #31, opened 2017-08-15
Add deckNamesToIds and deckIds actions

techdavid:
`deckNamesToIds` converts an array of names to IDs; `deckIds` returns the deck IDs in the same order as `deckNames` (useful for multi queries to get both at a time).

Maybe we should expose the internal `deckNameFromId` function as `deckIdsToNames`?

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -576,6 +576,14 @@ class AnkiBridge:
         self.stopEditing()
 
 
+    def deckNamesToIds(self, decks):
+        deckIds = []
+        for deck in decks:
+            id = self.collection().decks.id(deck)
+            deckIds.append(id)
+        return deckIds
+
+
     def cardsToNotes(self, cards):
         return self.collection().db.list('select distinct nid from cards where id in ' + anki.utils.ids2str(cards))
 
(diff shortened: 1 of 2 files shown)


FooSoft:
So my issue with this change is that `deckIds` is never going to be useful by itself... you are always going to have to call `deckNames` and then cross-reference the results with the output of `deckIds`. The other problem is that this type of code assumes that IDs are going to be ordered the same way as the names, which IMO is not great.

I think the underlying problem is that the original `deckNames` method just doesn't return enough information. I suggest that you add a new API called `getDecks` (or something similar) that returns a dictionary keyed on deck name, with values being dictionaries containing fields like `id`. The good thing about this approach is that if we decided to extend the amount of information that AnkiConnect clients could get about decks we could easily do so without breaking the API or adding additional methods.

This proposal would of course create overlap with the `deckNames` function (making it obsolete), but at the same time it will remove the need to have `deckNamesToIds`.

techdavid:
Ok I'll get to work shortly - to be honest I was thinking about doing 
something like this initially but decided against it because of the 
overlap!

PR #32, opened 2017-08-16
Add guiStartCardTimer function

CharlesHenry:
Starts or resets the 'timerStarted' value for the current card. This is useful for deferring the start time to when it is displayed via the API, allowing the recorded time taken to answer the card to be more accurate when calling guiAnswerCard.


FooSoft:
Looks good, thanks for the contribution 👍

PR #34, opened 2017-08-17
Add config group actions

techdavid:
I also moved the multi action to what I think is a more sensible position than the first place I first put it - if you disagree with this I'll happily change it back.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -509,7 +509,7 @@ class AnkiBridge:
                 return [field['name'] for field in model['flds']]
 
 
-    def confForDeck(self, deck):
+    def getDeckConfig(self, deck):
         if not deck in self.deckNames():
             return False
 
@@ -517,7 +517,7 @@ class AnkiBridge:
         return self.collection().decks.confForDid(did)
 
 
-    def saveConf(self, conf):
+    def saveDeckConfig(self, conf):
         confId = str(conf['id'])
         if not confId in self.collection().decks.dconf:
             return False
@@ -533,7 +533,7 @@ class AnkiBridge:
         return True
 
 
-    def changeConf(self, decks, confId):
+    def setDeckConfigId(self, decks, confId):
         for deck in decks:
             if not deck in self.deckNames():
                 return False
@@ -548,7 +548,7 @@ class AnkiBridge:
         return True
 
 
-    def addConf(self, name, cloneFrom=1):
+    def cloneDeckConfigId(self, name, cloneFrom=1):
         if not str(cloneFrom) in self.collection().decks.dconf:
             return False
 
@@ -556,7 +556,7 @@ class AnkiBridge:
         return self.collection().decks.confId(name, cloneFrom)
 
 
-    def remConf(self, configId):
+    def removeDeckConfigId(self, configId):
         if configId == 1 or not str(configId) in self.collection().decks.dconf:
             return False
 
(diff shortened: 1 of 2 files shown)


FooSoft:
Couple of things about this request:

* Although I realize you are just copying the internal Anki method names, it is hard to discover what functions like `remConf` do without looking at the documentation. It also goes against the convention of not using abbreviations in API names. Also they are not always representative of what they do. Take a look at in-code markup I'm adding.
* I noticed that you use `id` in a couple of places. This is actually a python keyword and even though it works fine as a variable it looks kind of goofy with syntax highlighting, so it would be good to switch it to something else.

AnkiConnect.py
```diff
@@ -502,11 +509,59 @@ def modelFieldNames(self, modelName):
                 return [field['name'] for field in model['flds']]
 
 
-    def multi(self, actions):
-        response = []
-        for item in actions:
-            response.append(AnkiConnect.handler(ac, item))
-        return response
+    def confForDeck(self, deck):
```
FooSoft: Rename this to `getDeckSettings`

AnkiConnect.py
```diff
@@ -502,11 +509,59 @@ def modelFieldNames(self, modelName):
...
+
+    def addConf(self, name, cloneFrom=1):
+        if not str(cloneFrom) in self.collection().decks.dconf:
+            return False
+
+        cloneFrom = self.collection().decks.getConf(cloneFrom)
+        return self.collection().decks.confId(name, cloneFrom)
+
+
+    def remConf(self, id):
```
FooSoft: This should be `removeDeckConfig`

AnkiConnect.py
```diff
@@ -502,11 +509,59 @@ def modelFieldNames(self, modelName):
...
+
+        conf['mod'] = mod
+        conf['usn'] = usn
+
+        self.collection().decks.dconf[id] = conf
+        self.collection().decks.changed = True
+        return True
+
+
+    def changeConf(self, decks, confId):
```
FooSoft: How about `setDeckConfigId`

AnkiConnect.py
```diff
@@ -502,11 +509,59 @@ def modelFieldNames(self, modelName):
...
+            return False
+
+        for deck in decks:
+            did = str(self.collection().decks.id(deck))
+            aqt.mw.col.decks.decks[did]['conf'] = confId
+
+        return True
+
+
+    def addConf(self, name, cloneFrom=1):
```
FooSoft: This should be `cloneDeckConfigId`

AnkiConnect.py
```diff
@@ -502,11 +509,59 @@ def modelFieldNames(self, modelName):
...
-        return response
+    def confForDeck(self, deck):
+        if not deck in self.deckNames():
+            return False
+
+        id = self.collection().decks.id(deck)
+        return self.collection().decks.confForDid(id)
+
+
+    def saveConf(self, conf):
```
FooSoft: What is this being used for? It doesn't seem like there are any APIs to actually modify the configuration profiles, so why do we want to save them?

techdavid: Use `getDeckConfig` to get the object, then modify it and save it again. This is an example in the project I'm writing:

image

This changes the config of the default deck so that the intervals for new cards are as you can see, and the max number of new cards is allowed (the GUI options window doesn't allow higher than 9999). When saving, no more parameters are needed because the config ID is included with the object.

FooSoft: Ah, I see. In that case it makes sense for it to exist. Let's call it `saveDeckConfig`.

AnkiConnect.py
techdavid: ID here is referring to the configuration ID, not deck ID

techdavid:
I've made the requested changes, can you check everything looks OK? (I'll check everything still works in my web app)

techdavid:
Forgot to update readme, I'll just do that first

techdavid:
I've tested all the actions - everything checks out. README is now updated.

FooSoft:
Thanks for making the changes -- looks good to me 👍

PR #35, opened 2017-08-18
Fix broken guiCurrentCard

techdavid:
A map can't be converted to JSON


FooSoft:
Ah yes, this doesn't work in Python 3 since `map` returns an iterator.

PR #37, opened 2017-08-20
Improve handling of HTTP headers

techdavid:
> I am not a fan of the direct indexing in `headers[2][1] = str(len(body))`

How does this look? As well as being nicer to read, this will also prevent duplicate headers.

Sorry for the trailing spaces I keep leaving! I switched to Sublime Text today and I've enabled removing them on save so it shouldn't happen any more.

How's the new version system going? If you're working on other stuff more right now I could have a go implementing what you described.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -202,16 +202,22 @@ class AjaxServer:
         self.resetHeaders()
 
 
-    def addHeader(self, name, value):
-        self.headers.append([name, value])
-    
-    
+    def setHeader(self, name, value):
+        self.headers[name] = value
+
+
     def resetHeaders(self):
-        self.headers = [
-            ['HTTP/1.1 200 OK', None],
-            ['Content-Type', 'text/json'],
-            ['Content-Length', '']
-        ]
+        self.headers = {
+            'HTTP/1.1 200 OK': None,
+            'Content-Type': 'text/json'
+        }
+
+
+    def getHeaders(self):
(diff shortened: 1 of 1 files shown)


FooSoft:
So while making the headers be a dictionary appears to be a natural way of doing things initially, it actually leads to undefined behavior in Python 3. I actually ran into this problem when porting AnkiConnect from Python 2 to 3. The issue is that ordering of items in a dictionary is not guaranteed in Python 3 -- this would sometimes cause the `HTTP 200 OK`  to come after the `Content-Type` and then everything goes to hell.

Now that I think about it, the proper way to fix this would be to leave the header array as it is , but make the `setHeader` method only modify an set of *optional* headers. So regardless of what the user sets, you always send the HTTP string, Content Type and Content Length. Afterwards you enumerate through any other optional headers and send those.

Regarding the AnkiConnect versioning, I've been pretty busy over the past couple of days with Yomichan but I have a pretty clear view of what I expect from the system. I'll probably be able to get around it next week now that Yomichan has been updated and it appears stable.

techdavid:
I've updated it to maintain the order for standard headers (i.e. 200 OK and Content-Type) - does this look ok?

techdavid:
don't merge yet, just tested and it isn't working currently

techdavid:
Tested and it's now working.

FooSoft:
Looks good :+1:

PR #39, opened 2017-08-21
Add storage actions

techdavid:
Reason for adding: I want to store settings for my app - like fields to test on for each note, testing options etc - in a way that it's saved between different browsers or even computers. By storing the JSON in the media folder, it's synced between different devices via AnkiWeb.

The underscore is required at the beginning of the filename to whitelist it, as otherwise Anki would otherwise warn that it's not used by any cards.

Should I try organizing the README? Now that we have lots more actions, I think it would be a good idea to separate them - e.g. a "Misc" header for stuff like `multi` and `version`, a "GUI" header for the GUI actions, a "Decks" header for `deckNames`, `deckNamesAndIds`, `getDeckConfig` etc...

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -17,6 +17,7 @@
 
 import anki
 import aqt
+import base64
 import hashlib
 import inspect
 import json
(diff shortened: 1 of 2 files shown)


FooSoft:
So I understand the reasoning behind what you are trying to do, but I don't think that it make a lot of sense when you think about it from an API design. Your application has a need to synchronize data between different computers, and for you that happens to be key-value pairs in JSON format. Now let's say that some other application came along and now it wanted to store binary data or lots of sparse files for what *it* was trying to do. What would end up happening over time is that the API would become full of application-specific behavior that is not globally useful. 

If you want to have an API that works on persistent files, why not do something where you actually manipulate `collection.media`, allowing files to be added and removed? This way people can use the API to add multimedia to collections in a generic way. You could also leverage the system to store JSON (or whatever else you wanted to). The data would all have to be base-64 encoded while being transferred across the API, but that is easy to do both on the Python and JavaScript ends.

FooSoft:
Regarding the `README.md` organization, it would indeed be good to organize things using headings so that developers could jump to the sections they care about without reading everything.

techdavid:
How does this look?

FooSoft:
Two things:
* All of the APIs should have `media` in them so for example `storeFile` becomes `storeMediaFile`.
* You should use the existing Anki APIs for working with media instead of doing the file I/O yourself. Look at how I add audio files for an example: `self.media().writeData(params.audio.filename, data)`

techdavid:
As far as I can tell, the Anki media API doesn't have support for directly reading or removing media (except for the aqt.mw.deleteUnused, as used in the "Check Media" dialog) - am I missing something? If not, should I still perform these operations manually?

techdavid:
I can't find an API for directly reading media - I suppose I should keep doing this manually? Or am I missing something?

FooSoft:
As far as reading media is concerned you are fine doing what you are doing right now. I imagine that one of the reasons that Anki does not have explicit APIs for reading data is that in many cases it uses external applications (such as `mplayer`) to load the multimedia files.

techdavid:
writeData creates a new file with the suffix (1), (2) etc. when trying to write to an existing file with different contents. I suppose this isn't desirable for an API so should I stick to doing this manually too?

FooSoft:
The main goal is ensuring that the media database is aware of the existence of the file that is being written out. The built-in file unduplication behavior in Anki can be avoided by executing the same code that does the file delete before doing the write.

techdavid:
How's the latest commit?

FooSoft:
Looks good, thanks for the changes 👍

Issue #40, opened 2017-08-28
Allow connections from other ip addresses


tomasgodoi:
Is it a security issue to allow connections only from 127.0.0.1?
On a docker environment, I needed to change NET_ADDRESS to 0.0.0.0. (At least to access it from a Windows environment with Docker Toolbox)
If it's not a security issue I could make a PR to change that.

FooSoft:
Hypothetically, since there is no authentication people could manipulate people's decks and Anki settings remotely but I don't see this actually ever happening in practice.  There are (by design) no commands to execute arbitrary code, so I don't see the point of someone writing such an exploit. I'm fine with your proposed change.

techdavid:
Tomas: You can use the undocumented internal function `setHeader` to allow access from a certain address, e.g. to allow `http://127.0.0.1:5678`:

    from AnkiConnect import ac
    ac.server.setHeader("Access-Control-Allow-Origin", "http://127.0.0.1:5678")

Or to allow `https://mysite.com`, just set that as the value instead.

Alex: I think it's a good idea to keep AC secure just in case - I found an exploit in Anki (now fixed) to allow executing arbitrary python code from importing a crafted deck, and I wouldn't be surprised if there are other bugs like this.

Also, shall I create a README section for internal functions designed to be used with other add-ons?

FooSoft:
@techdavid I guess you are right about the possibility of bugs in Anki itself being used for exploits. That would be pretty bad.

I'm against documenting the internals of AnkiConnect for the simple reason that if people started relying too much on its structure we would get into a situation where it would be impossible to change anything without potentially breaking someone's addon. I'm currently working on adding proper API versioning support in the `dev` branch, and there is no way I want to extend that coverage to internals.

For @tomasgodoi 's original use case it might be appropriate to just look for an environment variable called something `ANKICONNECT_BIND_ADDRESS`. If this environment variable does not exist then it would default to `127.0.0.1`, otherwise it would use the value of the environment variable.

tomasgodoi:
@FooSoft . I liked that approach. I'll make the PR.

tomasgodoi:
@techdavid, that `Access-Control-Allow-Origin` approach would work only if the server was accessible on the first place (to circumvent CORS restrictions). If the binding is made only on 127.0.0.1, the socket wouldn't be accessible from another host, so there's no chance to set that response header.
[...]

PR #41, opened 2017-08-28
Allowing to bind on other IP Addresses through an environment variable

tomasgodoi:
Fixes https://github.com/FooSoft/anki-connect/issues/40


FooSoft:
Nicely done 👍

Issue #42, opened 2017-08-28
Request headers should be case insensitive per RFC2616


tomasgodoi:
In fact, there's an assumption on AnkiConnect that the `Content-Length` header is on camel case.
That will break the behavior when you use a lib which doesn't write the headers like that.
That is the case when you use the node `request` (https://github.com/request/request) lib to call AnkiConnect.
I've wiresharked it and compared with a request made with Postman and found out it was the `Content-Length` header's fault, which was being sent as `content-length`.
AnkiConnect will use 0 as body length, `bodyLength = int(headers.get(makeBytes('Content-Length'), 0))`  (https://github.com/FooSoft/anki-connect/blob/master/AnkiConnect.py#L186) and will then return the default string `AnkiConnect v.4` (due to `if len(req.body) == 0:` on https://github.com/FooSoft/anki-connect/blob/master/AnkiConnect.py#L259)
I'm already working on a Pull Request to make them case insensitive.

PR #43, opened 2017-08-28
Making the incoming headers case insensitive.

tomasgodoi:
Any reference to incoming headers should now be made with lowercase.

Fixes https://github.com/FooSoft/anki-connect/issues/42.


FooSoft:
Looks good, thanks for the fix

PR #60, opened 2017-12-22
Add a header for CORS support

bertoncelj1:
In order to access this API from any page in the browser this header field is necessary or else the request gets blocked by CORS (Cross-Origin Resource Sharing) policy.

Having CORS enabled brings no additional security issues for this API, so there is not reason for it not to be enabled.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -217,7 +217,8 @@ class AjaxServer:
     def resetHeaders(self):
         self.headers = [
             ['HTTP/1.1 200 OK', None],
-            ['Content-Type', 'text/json']
+            ['Content-Type', 'text/json'],
+            ['Access-Control-Allow-Origin', '*']
         ]
         self.extraHeaders = {}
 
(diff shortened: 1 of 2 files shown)


FooSoft:
I wouldn't mind having having CORS be an option (that is off by default), but having it enabled for everyone is potentially dangerous. While I agree that there is nothing outwardly dangerous about the API, I don't think anyone can guarantee that there is no way to exploit it to gain arbitrary code execution through Anki. It would also be kind of bad if drive-by deck modification was possible through JavaScript on random pages on the internet.

bertoncelj1:
I understand your concerns, but having CORS does not prevent somebody to attack your app the way you described. Requests to the API are all POST so the only thing CORS prevents is for user to see content. If a attacker guesses deck names or note ids it can still manipulate them. The same goes for arbitrary code execution through Anki.  

Instead of relying on CORS i will probably write extension or use tampermonkey which are not limited as hard.  

Having CORS only as an option would also be an possibility. I just don't know where it would be possible to put this option inside an app. Do you have any suggestions?

FooSoft:
Sorry for the slow response; you have a point, I'll just take your request as is. It will be on the Anki page shortly. Thanks for your contribution!

PR #61, opened 2018-01-01
add support to modify fields of exist note.

jingtaozf:
To change fields directly without creating a same note.


FooSoft:
Looks good, thanks for the change!

Issue #51, opened 2017-09-22
Get information of any card


kgoedert:
Hi, 

It would be nice  to get information about any card (outside of review mode).

tomasgodoi:
It's already supported:
https://github.com/FooSoft/anki-connect#finding-notes-and-cards

tomasgodoi:
Maybe what you want is to retrieve the note fields? That would be useful to me too.

kgoedert:
Yes. I need a specific field on a card

tomasgodoi:
Well, maybe I can make a PR in the future if nobody does first.

ReneBrals:
This is relatively easy to do: the `guiCurrentCard` endpoint already retrieves card data, so a new endpoint can just adapt that functionality to fetch specific card IDs instead of the currently reviewed card. Doing this for notes works in a similar way.

I have a Pull Request open for this.

PR #63, opened 2018-01-09
Endpoint to get card/note fields and other data

ReneBrals:
Adds two endpoints, which allow the user to retrieve the fields and other data of a list of cards or notes. I needed this feature myself to write a script that fetches data through AnkiConnect, and there's the open issue #51 requesting it, so I thought I might as well put in a PR.

I added only the card data that is relatively easy to use. Specifically, I didn't add card due dates, because Anki stores those internally in a kind of weird way and I don't know for sure if it's useful to report it like that (a next suggestion is maybe to add an endpoint to give a card's due date as a proper date stamp?).

The responses are mostly formatted like the existing `guiCurrentCard` endpoint, because its functionality is similar.

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -762,7 +762,8 @@ class AnkiBridge:
                     'tags' : note.tags,
                     'fields': fields,
                     'modelName': model['name'],
-                    'cards': note.cards()
+                    'cards': self.collection().db.list(
+                        "select id from cards where nid = ? order by ord", self.id)
                 })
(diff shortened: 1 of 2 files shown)


FooSoft:
Looks good -- could you also update the `README.md` file please with notes about this API?

ReneBrals:
Done, they're in the README now.

FooSoft:
Merged, thanks! I will upload this to AnkiWeb shortly.

PR #69, opened 2018-02-21
Expose Anki sync via AnkiConnect

sudarshang:
I tested this code on the command line with

```
curl localhost:8765 -X POST -d "{\"action\": \"sync\", \"version\": 5}"
```

When sync is thus invoked, the Anki app window becomes active (moves above other windows on the desktop).  After invoking sync when the Anki app is closed the collection corrupt error is thrown. See [here](https://github.com/dae/anki/blob/master/aqt/main.py#L384)

I am working with the Anki21 beta build

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -948,7 +948,7 @@ class AnkiBridge:
         timer.start(1000) # 1s should be enough to allow the response to be sent.
 
     def sync(self):
-        self.window()._sync()
+        self.window().onSync()
 
 #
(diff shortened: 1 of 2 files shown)


sudarshang:
The corruption issue resolved after switching to onSync instead of _sync.

FooSoft:
Nice! Can you also update `README.md` please?

FooSoft:
Thanks for the code -- I'll push this to AnkiWeb in a little bit.

Issue #70, opened 2018-02-22
Is there ability to create a new empty deck?


c-okelly:
Apologies in advance if I have missed something in the docs but is there a specific action to create an empty deck? Or is there a reason that this functionality is not supported?

Thanks for releasing anki-connect!

FooSoft:
You are correct in that there is currently no way to create an empty deck. There is no particular reason for this feature not existing, it's just that nobody has added it. I mostly created AnkiConnect for the Yomichan extension, and people have been sending pull requests for APIs that they want to use. 

You're welcome to send a pull request with this feature, it's probably just a couple of lines of code :+1:

c-okelly:
Ahh cool. Thanks for the quick reply!

I will have a look at it!

FooSoft:
Yeah just look at the Anki source in the areas where the AnkiConnect plugin pokes around in it for deck operations. It should be pretty simple to do I think.

c-okelly:
Just opened a pull request there. Let me know if there is anything you would like changed!

FooSoft:
Merged :+1:

c-okelly:
Thanks for the quick comments, review and all the work in the first place!

FooSoft:
Yup, feel free to propose or implement any other API features you need!

PR #71, opened 2018-02-22
Add createDeck action

c-okelly:
Add action to create an empty deck. #70


FooSoft:
Looks good -- I'll update the AnkiWeb version later today.

Issue #72, opened 2018-03-08
Errors are only reported on an exception


c-okelly:
Hey,

I was using the api end point addNotes / addNote to insert new notes into the system and a noticed that there are a number of use cases where the note would fail to be created and the result would be None. There was no specific error message though as this two returned none.

Looking over the code I noticed that you only seem to return an error messages when an exception is thrown and caught. What would you think about refactoring the code to to include error messages when you have a logical failure.

An example of this might => attempt to add a new note. Deck specified doesn't exist. 
Results: None, Error: "The deck for specified note does note exist"

Just a suggestion anyway! 

Thanks,

C

FooSoft:
This is a good point. The exception handling logic you are talking about is a remnant of a time when errors could not be reported via the API. It should be trivial to make them properly get reported to users of the API.

c-okelly:
Ah makes sense! I would propose we create a custom exception class that could be used to throw an "action failed" exception with a more detailed error message. 

This could be caught by the existing code with no changes and the error message passed to back to the user?

It would also mean exceptions could be added incrementally. Let me know what you think!

C

FooSoft:
I don't think we even need a custom exception class, you can just do something as simple as `raise Exception('everything is broken');` and it would work. Any and all exceptions are automatically caught by the `handler` method and stuffed into the `error` member. Basically the only thing that needs to happen is that exceptions need to be thrown when there are problems with data provided by the API.

c-okelly:
Sounds good! I will work on something weekend and make a pull request.

PR #75, opened 2018-03-11
Extend error reporting when creating note.

c-okelly:
In reference to #72.

If the user adds an invalid note to Anki anki-connect will return a more specific error message.

Added a string method for class AnkiNoteParams.

Let me know if you want anything changed.

C

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -89,11 +89,11 @@ def makeStr(data):
 def download(url):
     try:
         resp = web.urlopen(url, timeout=URL_TIMEOUT)
-    except web.URLError:
-        return None
+    except web.URLError as e:
+        raise Exception("A urlError has occoured for url " + url + ". Error messages was: " + e.message)
 
     if resp.code != 200:
-        return None
+        raise Exception("Return conde for url request" + url + "was not 200. Error code: " + resp.code)
 
     return resp.read()
 
@@ -370,11 +370,11 @@ class AnkiBridge:
     def addNote(self, params):
         collection = self.collection()
         if collection is None:
-            return
+            raise Exception("Collection was not found.")
 
         note = self.createNote(params)
         if note is None:
-            return
+            raise Exception("Failed to create note from params" + str(params))
 
         if params.audio is not None and len(params.audio.fields) > 0:
             data = download(params.audio.url)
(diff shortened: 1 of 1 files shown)


AnkiConnect.py
```diff
@@ -89,11 +89,11 @@ def makeStr(data):
 def download(url):
     try:
         resp = web.urlopen(url, timeout=URL_TIMEOUT)
-    except web.URLError:
-        return None
+    except web.URLError as e:
+        raise Exception("A urlError has occoured for url " + url + ". Error messages was: " + e.message)
```
FooSoft: Single quotes instead of double quotes please (all throughout this PR)

AnkiConnect.py
```diff
@@ -89,11 +89,11 @@ def makeStr(data):
...
 
     if resp.code != 200:
-        return None
+        raise Exception("Return conde for url request" + url + "was not 200. Error code: " + resp.code)
```
FooSoft: conde => code

AnkiConnect.py
```diff
@@ -1110,11 +1118,14 @@ def addNote(self, note):
...
-            else:
-                results.append(None)
+            try:
+                params = AnkiNoteParams(note)
+                if params.validate():
+                    results.append(self.anki.addNote(params))
+                else:
+                    results.append(None)
+            except Exception as e:
+                results.append(str(e))
```
FooSoft: This change would mean that `addNotes` can return a note ID, None or a string. Existing code in things that use AnkiConnect likely have code that checks to see if result `is not None`, and this would break their behavior.

AnkiConnect.py
c-okelly: Will do.

c-okelly:
Will change it to return just none.

c-okelly:
I defiantly agree that line 1127 by introducing string return could break functionality. I have changed it so that the when an exception occurs it just append an None to the array.

The code I have added done on line 428-435 does mean now that if the card was a duplicate or an empty card it will throw an error. I would be concerned this might be more hassle then a benefit if people use the api to add a lot of cards in one go if the exception is not caught but it does also mean the information is buried. 

If you think it should be removed will do or if there are any other changes you think could be me let me know.

Thanks

AnkiConnect.py
c-okelly: Changed to append(None) instead

AnkiConnect.py
c-okelly: Corrected

FooSoft:
Thanks for the changes. I see your point about throwing exceptions half way through an operation, but I don't really think there is any other good way to report multiple errors through the API.

c-okelly:
No neither do it as it's is currently tbh. Thanks for the feedback though.

PR #77, opened 2018-03-16
Catch exception when checking if a note can be created

aci2n:
Regular user with Yomichan here.
Noticed that the "+" button would not show if a card in the "canAddNotes" query already existed, and it seems to have been due to some exceptions that were added some time earlier in AnkiBridge.createNote().
Just made it so that AnkiBridge.canAddNote() catches any exception thrown there and returns False in those cases.


FooSoft:
Nice, thanks for the fix!

PR #79, opened 2018-05-18
Add template field to guiCurrentCard output + increment API_VERSION

CharlesHenry:
Adding template to allow for calculation of which fields to show for each side of the card and in which order, regardless of model.

I've had to increment the API_VERSION on my side as there is no way of telling if this field is supported until guiCurrentCard is actually called.

I've noticed unit tests have been developed recently. Please let me know if modifications are required on my side to support this.


FooSoft:
Clients of the version 5 API will see no disruption when the `API_VERSION` is increased so there are no problems with this whatsoever. 

This change looks good, but please update `README.md` and the unit tests to use the new version number. Also you should update the sample response from `guiCurrentCard` so that other users are aware of the new `template` parameter and what it contains.

CharlesHenry:
Okie dokie I'll get onto this ASAP. Thanks!

CharlesHenry:
How's it looking now Alex? Let me know if I've missed anything.

FooSoft:
Looks good :+1:

PR #88, opened 2018-07-29
Convert iterator to list to make it JSON serializable.

### LINKED ISSUE #85: <class 'TypeError'>: Object of type 'map' is not JSON serializable
I tried with beta 42 and the latest AnkiConnect from 2018-06-24. This is the only addon installed. I don't call AnkiConnect directly but from within an emacs package that's called [anki-editor](https://github.com/louietan/anki-editor). This package allows to send/push notes to AnkiConnect. When I run the "push" command in this package I get the following error message in Anki. 

This emacs package works fine with AnkiConnect in 2.0.52. So I assume that the problem is not with this emacs-package? I don't know about the internals of emacs/elisp so I can't tell you what exactly is sent to AnkiConnect. Nevertheless I decided to post because this error message might still be useful for you ? If you feel that a novice like me can help with further information just give me instructions on what to do. 

Anyway thanks for your great software. 

```

Debug info:
Anki 2.1.0beta42 Python 3.6.2 Qt 5.9.2 PyQt 5.9
Platform: Linux
Flags: frz=True ao=True sv=1

Caught exception:
  File "/the/path/to/profile/addons21/2055492159/__init__.py", line 293, in advance
    self.server.advance()
  File "/the/path/to/profile/addons21/2055492159/__init__.py", line 204, in advance
    self.advanceClients()
  File "/the/path/to/profile/addons21/2055492159/__init__.py", line 219, in advanceClients
    self.clients = list(filter(lambda c: c.advance(), self.clients))
  File "/the/path/to/profile/addons21/2055492159/__init__.py", line 219, in <lambda>
    self.clients = list(filter(lambda c: c.advance(), self.clients))
  File "/the/path/to/profile/addons21/2055492159/__init__.py", line 126, in advance
    self.writeBuff += self.handler(req)
  File "/the/path/to/profile/addons21/2055492159/__init__.py", line 238, in handlerWrapper
    body = makeBytes(json.dumps(self.handler(params)))
  File "json/__init__.py", line 231, in dumps
  File "json/encoder.py", line 199, in encode
  File "json/encoder.py", line 257, in iterencode
  File "json/encoder.py", line 180, in default
<class 'TypeError'>: Object of type 'map' is not JSON serializable
```

louietan:
This fixes #85

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -489,7 +489,7 @@ class AnkiConnect:
 
     @api()
     def multi(self, actions):
-        return map(self.handler, actions)
+        return list(map(self.handler, actions))
 
 
     #


FooSoft:
Ah, so it's a Python 3 thing... Thanks for the fix!
Will upload to AnkiWeb shortly.

PR #97, opened 2018-10-19
Use AnkiRequestsClient for downloads in Anki 2.1

### LINKED ISSUE #96: HTTPS downloading broken in Anki 2.1 Mac/Linux
The exception is `<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:748)>`

I'm using the Anki 2.1.5 binary release on Fedora 28. Other reports: https://github.com/FooSoft/yomichan/issues/137

It works for me in source or binary 2.0.52 on Fedora; both versions on Windows 7; and if I run the relevant AnkiConnect code in system python3 on Fedora (so the issue is probably in Anki's bundled libraries, and it would probably work if I was running Anki 2.1 from source).

I think I have a good solution so will post that shortly.

HelenFoster:
Fixes #96

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -53,15 +53,27 @@ URL_UPGRADE = 'https://raw.githubusercontent.com/FooSoft/anki-connect/master/Ank
 
 if sys.version_info[0] < 3:
     import urllib2
-    web = urllib2
+    def download(url):
+        contents = None
+        resp = urllib2.urlopen(url, timeout=URL_TIMEOUT)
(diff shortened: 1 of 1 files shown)


FooSoft:
Looks good, I assume that `AnkiRequestsClient` is not available for old versions of Anki (thus requiring the code branch)?

HelenFoster:
Right, `AnkiRequestsClient` doesn't exist in the 2.0 versions.

Issue #98, opened 2018-11-04
addNote should support duplicate notes...


burtonator:
If I have a flashcard with the same field you prevent it from being added BUT the anki client itself can add it just fine as a duplicate card?

Duplicate cards are SUPPORTED just not recommended but anki connect blocks them.

I get the message "cannot create note because it is a duplicate"

I suggest adding a param of "allow_duplicates"

alecschule:
Perhaps there should be a separate action for adding new cards? (to an existing note) In any case, something along these lines would be useful to me as well.

bertoncelj1:
Instead of adding a new action I would suggest adding new field in note group called "options" where user could define parameters such as "allow_empty", "allow_duplicate" ... 

```
{
    "action": "addNotes",
    "version": 6,
    "params": {
        "notes": [
            {
                "deckName": "Default",
                "modelName": "Basic",
                "fields": {
                    "Front": "front content",
                    "Back": "back content"
                },
                "tags": [ "yomichan" ],
                "options": {
                     "allowDuplicate": true, 
                }
            }
        ]
    }
}
```

burtonator:
Heck yeah.. that would be perfect for me.

bertoncelj1:
Just out of curiosity, what are you developing that you need this feature?

FooSoft:
The associated pull request has been merged and an updated version of the plugin has been uploaded to AnkiWeb :+1:

burtonator:
WOW.. nice.  

I'm going to add this to the next version of Polar :) 

https://getpolarized.io/

PR #99, opened 2018-11-07
Add option to allow adding duplicate notes

bertoncelj1:
Added a feature to allow adding duplicate notes. As requested here: https://github.com/FooSoft/anki-connect/issues/98

Duplicate notes can be added by enabling option inside a note like this:

```
{
    "action": "addNote",
    "version": 5,
    "params": {
        "note": {
            "deckName": "TestDeck",
            "modelName": "Basic",
            "fields":  ... ,
            "options": {
            	"allowDuplicate": true
            },
            "tags": ... ,
            "audio": ...
        }
    }
}
```

I also added a brief description of this option in README

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -449,11 +449,21 @@ class AnkiConnect:
             if name in ankiNote:
                 ankiNote[name] = value
 
+        allowDuplicate = False
+        if 'options' in note:
+          if 'allowDuplicate' in note['options']:
+            allowDuplicate = note['options']['allowDuplicate']
+            if type(allowDuplicate) is not bool:
+              raise Exception('option parameter \'allowDuplicate\' must be boolean')
+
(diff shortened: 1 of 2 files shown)


AnkiConnect.py
```diff
@@ -449,11 +449,21 @@ def createNote(self, note):
             if name in ankiNote:
                 ankiNote[name] = value
 
+        allowDuplicate = False
+        if "options" in note:
```
FooSoft: Single quotes for strings please

AnkiConnect.py
```diff
@@ -449,11 +449,21 @@ def createNote(self, note):
...
+          if  "allowDuplicate" in note["options"]:
```
FooSoft: Looks like a stray space after the `if` here ; )

PR #101, opened 2018-12-01
Add API endpoint to load a profile

z1lc:
If this needs tests please let me know

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -509,19 +509,18 @@ class AnkiConnect:
 
     @api()
     def loadProfile(self, name):
+        if name not in self.window().pm.profiles():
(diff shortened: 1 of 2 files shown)


z1lc:
* Changed to return `True`/`False`
* Handles invalid profile names by returning `False`
* Updated README with example use

FooSoft:
Looks good :+1:

bikenik:
Hello. Could you add the endpoint to return current profile name? I've got several profiles and I toggle them all the time. I need it to dynamic change the path in my app to `collection.media` folder. Hope it should be useful for others also.

```py
@api()
def loadProfileName(self):
    return self.window().pm.name
```

Issue #10, opened 2017-06-05
Deck and note type parameters for guiAddCards


Profpatsch:
It would be nice to be able to open the GUI entry window with the correct deck/note type already preselected, e.g. for creating a keyboard shortcut:

```bash
function addCard {
  local deck=$1
  local note=$2
  # from package `httpie`
  http --json POST localhost:8765 action=guiAddCards 'params={"deck": "$deck", "note_type=$note"}'
}

# add these to your keybindings
addCard Japanese "Japanese (recall)"
addCard English Basic
```

something like that.

hgiesel:
It would be even nicer to prefill fields with content that you can send over the API. I wanna take a look into this in the coming weeks...

hgiesel:
I think this can also be closed.

Profpatsch:
@hgiesel Oh cool, this can be done now I assume?

hgiesel:
Yes. Look at the documentation for `guiAddCards`.

PR #103, opened 2019-01-17
Extend guiAddCards so that it's possible to preset notes

hgiesel:
This would add much more functionality to the guiAddCards endpoint. It basically gives it all the options that addNote currently has (except the sound parts, but I'm sure it wouldn't be hard to add that as well).
For more information, also see #10

I've also added something a bit more out there:
Enabling the option 'closeAfterAdding', you can create an `AddCards` window which closes as soon as you create the note. I find it very helpful, if you technically have all the card information you want, but you want to give the user the ability to review the information before commiting it to the collection.
Makes it also more reasonable to open multiple windows, which would all close, as you accept/edit the cards, instead of remaining on the screen.


hgiesel:
I noticed two more things:
* Creating two "AddAndClose" dialogs, closing one, and exiting Anki crashes. The reason is that every "AddAndClose" dialog signs up on the windowing system with the same name.
* You cannot simply open a dialogue with just "option", but not "modelName", or "deckName", but I see that more as a feature than a bug

hgiesel:
Ok thanks for the feedback.
Another thing I want to mention: even if calling `guiAddCards` with a note parameter will recreate the window, I will leave the call without the `note` parameter to just _activate the window_, keeping backwards compatibility in that sense. Basically there'll be three distinct use cases for the `guiAddCards` action: Activating the window, opening/reopening the `AddCards` dialog, opening `AddCardsAndClose` dialog. I'll try to reflect that in the README as well.

hgiesel:
Okay I fixed all the bugs, created unit tests, and updated the README.
The only small caveat: On Anki 2.0, it won't recognize the new content of the fields, until you click on them once. But I made sure, the closeAfterAdding option doesn't close the window until you actually added something...

hgiesel:
I have to admit I found another bug concerning model ids...
When I have the time I'll try and fix it...

hgiesel:
OK, I fixed all bugs that were still present. However the caveat on Anki 2.0 previously mentioned still stands. Besides that, everything works as intended.

FooSoft:
Looks good, but what is `meta.json`?

hgiesel:
Oops, the file must have been created by Anki without me noticing. I removed it.

FooSoft:
Awesome, thanks for the code!

PR #107, opened 2019-02-18
Bug fix, duplicate audio filenames

derxe:
Fix the bug where old file name was referenced in the audio field. 

The problem was, if the audio file we were trying to save already existed it got automatically saved with a different name (the number was appended to the end), but inside the card the incorrect old filename was referenced.


FooSoft:
Looks good, thanks for the code!

Issue #100, opened 2018-11-07
Ability to create custom note types.


burtonator:
I'd like the ability to add custom note types.  

With Polar I want the ability to add new fields like Author, URL, screenshot, so that when I sync over notes they have extended metadata.

Right now I can only use the pre-existing note types ...

FooSoft:
AnkiConnect was built as something to be used alongside Anki, so the idea is that the user sets up their deck there, and then your application allows the user to specify how they want the fields populated. So yes, you can already use custom note types. If you want to customize models in some way, I wouldn't mind a pull request with that behavior.

c-okelly:
I too would be interested in this and might have the time to work on it.

Main issue would be to choose a proposed format for how to create the new models while and ensure as much flexibility as possible. I tried to choose the names to be as verbose as possible. Card templates can be 1-n. Fields also 1-n (Though not sure about the single field use case? Maybe 2-n?)

``` json
{
    "action": "createNewModel",
    "version": 6,
    "params": {
        "modelName": "newModelName",
        "inOrderFields": ["Field1", "Field2", "Field3"], 
        "cardTemplates": [
            {
                "Front": "Front html",
                "Back": "Back html",
                "Css": "Optional CSS and default to builtin css",
            }
        ]
    }
}
```

Let me know what you think anyway

c-okelly:
@FooSoft I think this issue can probably be closed!

crimx:
Certain names(Date, Text, Note...) in `inOrderFields` are translated to other language if Anki is not in English. How to prevent that?

crimx:
I think this is a bug since the field names in card template are not translated.

PR #108, opened 2019-02-22
Add new createModel Api end point without documentation

c-okelly:
Pull request for issue #100 .

I am opening this pull request for @FooSoft review. If you like the over all format I will update it with documentation for the readme.

If you want any overall changes let me know.

Code was mainly built using:
        https://github.com/dae/anki/blob/b06b70f7214fb1f2ce33ba06d2b095384b81f874/anki/stdmodels.py

As referenced in my comment api endpoint is as follow:

inOrderField: (1-n) (Do we want to throw an error if at least 1 value it not present?)
cardTemplate: (1-n) (Do we want to throw an error if at least 1 value it not present?)
``` json
{
    "action": "createNewModel",
    "version": 6,
    "params": {
        "modelName": "newModelName",
        "inOrderFields": ["Field1", "Field2", "Field3"], 
        "cardTemplates": [
            {
                "Front": "Front htm {{Field1}}l",
                "Back": "Back html  {{Field2}}",
                "Css": "Optional CSS and default to builtin css",
            }
        ]
    }
}
``` 

Questions:

- [ ] Overall API issues

- [ ] Throw error if correct numbers not present? (cardTemplates / inOrderFields)

- [ ] Does the back and front need to validate fields are actually used?

- [ ] Any formatting changes

- [ ] Default CSS options or not?

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -32,7 +32,6 @@ from operator import itemgetter
 from time import time
 from unicodedata import normalize
 
-
 #
 # Constants
 #
@@ -849,8 +848,15 @@ class AnkiConnect:
 
 
     @api()
-    def createModel(self, modelName, inOrderFields, cardTemplates):
+    def createModel(self, modelName, inOrderFields, cardTemplates, css = None):
         # https://github.com/dae/anki/blob/b06b70f7214fb1f2ce33ba06d2b095384b81f874/anki/stdmodels.py
+        if (len(inOrderFields) == 0):
+            raise Exception('Must provide at least one field for inOrderFields')
+        if (len(cardTemplates) == 0):
+            raise Exception('Must provide at least one card for cardTemplates')
+        if (modelName in self.collection().models.allNames()):
+            raise Exception('Model name already exists')
+
         collection = self.collection()
         mm = collection.models
 
@@ -861,8 +867,11 @@ class AnkiConnect:
         for field in inOrderFields:
             fm = mm.newField(_(field))
             mm.addField(m, fm)
+        
+        # Add shared css to model if exists. Use default otherwise
+        if (css is not None):
+            m['css'] = css
 
-        # TODO? => There is no validation that card fields are used correctly
         # Generate new card template(s)
         cardCount = 1
         for card in cardTemplates:
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -849,6 +849,34 @@ class AnkiConnect:
 
 
     @api()
+    def createModel(self, modelName, inOrderFields, cardTemplates):
+        # https://github.com/dae/anki/blob/b06b70f7214fb1f2ce33ba06d2b095384b81f874/anki/stdmodels.py
+        collection = self.collection()
+        mm = collection.models
+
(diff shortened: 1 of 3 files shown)


FooSoft:
Thank you for the pull request and doing it in the form of a discussion. It makes things much easier for me :+1:

The API looks nice and clean, I like it. I wouldn't worry too much about validation except the simple case of testing to see if the model already exists, since people are likely to just call this API over and over. While additional validation is always good, it would add to the complexity to the extension (and it would be difficult to do thorough coverage for absolutely everything).

As for CSS, I think just having the "Css" field be optional is the way to go. If it's not provided, just use Anki defaults. 

Oh! Also, please add unit tests!

c-okelly:
Sorry only saw your comment today.

Would it be fair to say the following then for completion?

- [ ] Make CSS field optional and default
- [ ] ensure at least one card template (to avoid a generating incomplete cards )
- [ ] ensure at least one field provided (to avoid a generating incomplete cards )
- [ ] Check if model already exists
- [ ] Add unit test

FooSoft:
Yup, that would be what's left :+1:

c-okelly:
Sorry for the delay. It is all done now. 

I have added a test for adding with and without css but the test only really test that the Api call is working rather then assert any of the effects. Is that enough?

Turns out the CSS is tied to the model rather then the individual cards so you can only set the css once for a group of models.

It will throw an exception if the users add new models with no cards / fields.

Also docs have been updated

c-okelly:
The expected result is super long is the only issues not sure if you want it included or not?

```json

{
   "result":{
      "sortf":0,
      "did":1,
      "latexPre":"\\documentclass[12pt]{article}\n\\special{papersize=3in,5in}\n\\usepackage[utf8]{inputenc}\n\\usepackage{amssymb,amsmath}\n\\pagestyle{empty}\n\\setlength{\\parindent}{0in}\n\\begin{document}\n",
      "latexPost":"\\end{document}",
      "mod":1551462107,
      "usn":-1,
      "vers":[

      ],
      "type":0,
      "css":".card {\n font-family: arial;\n font-size: 20px;\n text-align: center;\n color: black;\n background-color: white;\n}\n",
      "name":"TestApiModel",
      "flds":[
         {
            "name":"Field1",
            "ord":0,
            "sticky":false,
            "rtl":false,
            "font":"Arial",
            "size":20,
            "media":[

            ]
         },
         {
            "name":"Field2",
            "ord":1,
            "sticky":false,
            "rtl":false,
            "font":"Arial",
            "size":20,
            "media":[

            ]
         }
      ],
      "tmpls":[
         {
            "name":"Card 1",
            "ord":0,
            "qfmt":"",
            "afmt":"This is the back of the card {{Field2}}",
            "did":null,
            "bqfmt":"",
            "bafmt":""
         }
      ],
      "tags":[

      ],
      "id":"1551462107104",
      "req":[
         [
            0,
            "none",
            [

            ]
         ]
      ]
   },
   "error":null
}
```

FooSoft:
Looks good, I don't mind the long JSON, tells people what they are going to see when they use the API. Thanks  for making the requested changes :+1:

Issue #111, opened 2019-03-04
API returns noteid even when no note is added. 


derxe:
Function addNote returns returns number of new cards added. If all the templates are blank no new cards are added and 0 is returned.

In [anki/collection.py](https://github.com/dae/anki/blob/ebdd8dae4b94327f83470ab6f72033fd81cb456f/anki/collection.py). line 298. 
```python
# check we have card models available, then save
cms = self.findTemplates(
if not cms:
    return 0
```
The API should check for that and return an exception in case when no new cards are added. 

An example how anki handles no new cards added. In[ aqt/addCards.py](https://github.com/dae/anki/blob/ebdd8dae4b94327f83470ab6f72033fd81cb456f/aqt/addcards.py) line 171 we have: 
```python
cards = self.mw.col.addNote(note)
if not cards:
    showWarning(_("The input you have provided would make an empty question on all cards."), help="AddItems")
    return
```

API does not check for this exception and returns note id as if the note was successfully added. 

Took me quite a while to find this bug :). Will create pull request fixing this bug shortly.

snoyes:
I reckon this bug can be closed.

FooSoft:
Ah yup, good call

PR #112, opened 2019-03-05
Fix bug where API returns noteid even when no note is added

derxe:
https://github.com/FooSoft/anki-connect/issues/111


FooSoft:
Looks good :+1:

PR #116, opened 2019-05-05
Fix README.md

dragonman225:
Hello, I found that there seems to be a typo in the documentation.
There is an API action "createNewModel" in the documentation, but it is not working.
Instead, after reading the code, I found out that the working action is "createModel".
So I changed "createNewModel" to "createModel" in the README.md.


FooSoft:
Nice catch!

PR #119, opened 2019-05-25
Add next review times when returning data in guiCurrentCard

corollari:
A user requested a feature in corollari/ankiTab#19 that would require being able to obtain the times for the next review assigned to each button. This pull request adds a new data field to guiCurrentCard that retrieves that information.

As nothing else is modified, this is backwards compatible with the current version of the API.


FooSoft:
Looks good, thanks :smile:

PR #124, opened 2019-09-22
Update Python example for Python 3

leezu:
Python 2 is deprecated and will be end-of-life by the end of this year https://pythonclock.org/


FooSoft:
Looks good to me :+1:

PR #125, opened 2019-10-27
Add get and update endpoints for model templates and styling

benkogan:
Includes basic tests and updates to README.md documentation.

I recently threw together a script for loading and modifying card templates and styling so that I could edit cards for my notes in a proper editor. To enable said script, I made the changes included in this PR to AnkiConnect locally and have been using them for month or two now without any significant issues.

(I say "significant" issues because there is (very infrequently) a single small issue where Anki will prompt me to check the database if I try updating in a certain state. I'm not sure what exact state that is, but iirc it tends happens when I do something like update templates while the Anki browser is open and I'm previewing the card I'm updating. It only happens infrequently and the only consequence is that after I check the database, Anki needs to do a full upload but no data is lost. Thus far it hasn't been an inconvenience, so I haven't pursued it further nor tried to find exact reproduction steps.)

Thank you for all your work on AnkiConnect — it's a wonderful add-on!

### DIFF
--- a/AnkiConnect.py
+++ b/AnkiConnect.py
@@ -947,6 +947,64 @@ class AnkiConnect:
 
         return templates
 
+    @api()
+    def modelTemplates(self, modelName):
+        model = self.collection().models.byName(modelName)
+        if model is None:
+            raise Exception('model was not found: {}'.format(modelName))
+
+        templates = {}
+        for template in model['tmpls']:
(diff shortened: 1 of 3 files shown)


FooSoft:
Sorry it took me a while to get to this, but good work!

FooSoft:
I will update the plugin on AnkiWeb with this change shortly

benkogan:
Perfect, thank you!

PR #128, opened 2019-11-26
Fix a couple issues with spelling/capitalization

zmughal:


FooSoft:
Thanks for the fixes!

PR #131, opened 2019-12-20
fix #130 - restrict cors to localhost

### LINKED ISSUE #130: Security: restrict CORS to localhost
Hi all,

Currently CORS allow all origins.
By default it would be better to either disable or restrict to localhost.

Ideally it should be configurable, otherwise any malicious website can mess up with your Anki and also your computer.

Any questions, please let me know.
Anderson

andersonDadario:
Details in the issue.


FooSoft:
Looks good, thanks for the fix!

PR #133, opened 2020-02-02
Ignore duplicates which are not in the target deck

toasted-nutbread:
Notes with the same content cannot be added to different decks since [`dupeOrEmpty`](https://github.com/ankitects/anki/blob/ab1f2429a77d1c366f4f4b44b0cb67ec95025f54/pylib/anki/notes.py#L184-L199) does not filter based on deck ID. I recently ran into this issue with Yomichan. Not sure if this is the correct way to handle this case.


FooSoft:
Looks good to me, thanks 👍

PR #134, opened 2020-02-16
Replace Card._getQA() with the new answer() and question()

qwhex:
I needed this to get the extension work with Anki 2.1.20.

Extra: add `link_mac.sh` and fix shebangs to work universally


FooSoft:
Looks good, thanks!

PR #136, opened 2020-02-17
Add support for multiple cors origins

mauamy:
Add support for configuring multiple cors origins by using the *origin* request-header and comparing its value to the configured cors origins.

### DIFF
--- a/plugin/config.json
+++ b/plugin/config.json
@@ -3,5 +3,5 @@
     "apiLogPath": null,
     "webBindAddress": "127.0.0.1",
     "webBindPort": 8765,
-    "webCorsOrigin": "http://localhost"
+    "webCorsOrigin": ["http://localhost"]
 }
--- a/plugin/config.json
+++ b/plugin/config.json
@@ -3,5 +3,6 @@
     "apiLogPath": null,
     "webBindAddress": "127.0.0.1",
     "webBindPort": 8765,
-    "webCorsOrigin": ["http://localhost"]
+    "webCorsOrigin": "http://localhost",
+    "webCorsOriginList": ["http://localhost"]
 }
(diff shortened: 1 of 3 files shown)


plugin/web.py
```diff
@@ -153,10 +153,20 @@ def handlerWrapper(self, req):
             except ValueError:
                 body = json.dumps(None).encode('utf-8')
 
+        # handle multiple cors origins by checking the 'origin'-header against the allowed origin list from the config
+        webCorsOriginsSetting = util.setting('webCorsOrigin')
+        corsOrigin = "http://localhost"
```
FooSoft: Single quotes please

plugin/config.json
```diff
@@ -3,5 +3,5 @@
     "apiLogPath": null,
     "webBindAddress": "127.0.0.1",
     "webBindPort": 8765,
-    "webCorsOrigin": "http://localhost"
+    "webCorsOrigin": ["http://localhost"]
```
FooSoft: Does changing the type of the setting value work fine for people who have installed the previous version? May be better to retire the current key and instead create one called webCorsOriginList or something like that.

mauamy: Yes, you're are right. I'll rename it accordingly. Do you propose to remove the old key immediately or just mark it as deprecated and keep the support until it will be removed in the future?

plugin/web.py
mauamy: Changed it.

plugin/config.json
mauamy: I have renamed the field to 'webCorsOriginList' like you proposed but kept support for the original 'webCorsOrigin' field (for now). I would like to remove it completely, because this leads to some unnecessary configuration redundancy. Do you see any related problems that might occur?

mauamy:
It seems like the *webCorsOriginList* field is not existent when downloading anki-connect via AnkiWeb, which breaks the connectability. I created issue #140 for this.

PR #138, opened 2020-02-23
Accept multiple audio files in note creation and update

andreaskienle:
Hi there,

First of all thank you so much for offering this plugin and the awesome work that you put into this.

This commit changes the API to accept an array of audio files instead of a single object.

Example:

``` json
{
    "action": "addNote",
    "version": 6,
    "params": {
        "note": {
            "audio": [{
                "url": "https://www.oxfordlearnersdictionaries.com/media/english/uk_pron/s/ser/seren/serendipity__gb_1.mp3",
                "filename": "oald_serendipity__gb_1.mp3",
                "fields": ["Audio"]
            }, {
                "url": "https://www.oxfordlearnersdictionaries.com/media/english/uk_pron/s/ser/seren/serendipity__us_2.mp3",
                "filename": "oald_serendipity__us_2.mp3",
                "fields": ["Audio"]
            }]
        }
    }
}
```

It also extends the API for the note update (`updateNoteFields`) to accept an array (or a single object) for audio file(s). Previously there was no support for adding audio after the note has been created.

I hope this is useful for other users too.

Let me know if I should document this in the readme.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -485,7 +485,7 @@ class AnkiConnect:
                                     ankiNote[field] += u'[sound:{}]'.format(audioFilename)
 
                     except Exception as e:
-                        errorMessage = str(e).replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
+                        errorMessage = str(e).replace('&', '&amp;').replace('<', '&lt;').replace('>', '&gt;')
                         for field in audio['fields']:
                             if field in ankiNote:
                                 ankiNote[field] += errorMessage
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -449,29 +449,8 @@ class AnkiConnect:
     def addNote(self, note):
         ankiNote = self.createNote(note)
 
-        audio = note.get('audio')
-        if audio is not None and len(audio['fields']) > 0:
-            try:
-                data = util.download(audio['url'])
-                skipHash = audio.get('skipHash')
-                if skipHash is None:
-                    skip = False
-                else:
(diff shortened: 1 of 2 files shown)


plugin/__init__.py
```diff
@@ -484,6 +463,34 @@ def addNote(self, note):
...
+                            skip = skipHash == m.hexdigest()
+
+                        if not skip:
+                            audioFilename = self.media().writeData(audio['filename'], data)
+                            for field in audio['fields']:
+                                if field in ankiNote:
+                                    ankiNote[field] += u'[sound:{}]'.format(audioFilename)
+
+                    except Exception as e:
+                        errorMessage = str(e).replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
```
FooSoft: Single quotes please

plugin/__init__.py
```diff
@@ -484,6 +463,34 @@ def addNote(self, note):
...
+                        for field in audio['fields']:
+                            if field in ankiNote:
+                                ankiNote[field] += errorMessage
```
FooSoft: Is there value in putting the error in the note? It seems more useful to just return error to caller?

andreaskienle: To be honest, I didn't really pay attention to this, I just extracted the code segment into the `addAudio` method without making any changes to the inner workings.

plugin/__init__.py
andreaskienle: done

andreaskienle:
> Can you also please update the README with the proposed change?

I've updated the README

PR #144, opened 2020-03-19
Supports profile specific configuration

zzzdeb:
I needed this, because my external program has also profile like feature and I wanted to distinguish them completely. This was specially useful when there are multiple anki instances (I forked anki and made some change).

Then I thought, there could be broader use case.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -60,9 +60,8 @@ class AnkiConnect:
             QMessageBox.critical(
                 self.window(),
                 'AnkiConnect',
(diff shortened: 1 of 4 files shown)


Kubo2:
@zzzdeb I am not completely sure what you are proposing, however it would be an improvement to be able to list and switch between profiles porgrammatically.

snoyes:
> @zzzdeb I am not completely sure what you are proposing, however it would be an improvement to be able to list and switch between profiles porgrammatically.

That feature has existed since https://github.com/FooSoft/anki-connect/pull/159

PR #150, opened 2020-04-04
Complete 975c069 fixes

corollari:
The fixes for old versions of anki that were introduced in 975c069 didn't cover `guiCurrentCard`, this PR extends them to cover it.


FooSoft:
Good work 👍

PR #151, opened 2020-04-04
Add REST interface for file retrieval

corollari:
When displaying cards outside of anki, getting the files embedded inside them to work properly has given me quite a lot of trouble, as right now this requires parsing the html document, requesting the files through the API and then embedding them into the document for display. This has been a constant source of problems due to the high variety of corner cases that appear when parsing, and the possibility of javascript being involved complicates everything to the extreme. 

This PR aims to solve that problem by providing another interface for retrieveMediaFile, one that essentially serves these files as a webserver. This will make it so html can be dumped on any browser-like system and the files will be automatically requested by that same browser.

**Note**: I'm not quite satisfied with the code quality of this PR, I think the conditionals I've introduced in web.py are a bit convoluted (I've spent like 2 hours trying to refactor but haven't come up with a better solution, suggestions are welcome!) and it's lacking some error handling, but I've submitted this PR to get the conversation started on this specific change.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -435,7 +435,7 @@ class AnkiConnect:
             raise Exception('You must either provide a "data" or a "url" field.')
 
     @util.api()
-    def retrieveMediaFile(self, filename):
+    def retrieveMediaFile(self, filename, encoding="base64"):
         filename = os.path.basename(filename)
         filename = unicodedata.normalize('NFC', filename)
         filename = self.media().stripIllegal(filename)
@@ -443,7 +443,7 @@ class AnkiConnect:
         path = os.path.join(self.media().dir(), filename)
         if os.path.exists(path):
             with open(path, 'rb') as file:
-                return base64.b64encode(file.read()).decode('ascii')
+                return base64.b64encode(file.read()).decode('ascii') if encoding == "base64" else file.read()
 
         return False
 
--- a/plugin/web.py
+++ b/plugin/web.py
@@ -145,7 +145,22 @@ class WebServer:
 
     def handlerWrapper(self, req):
         if len(req.body) == 0:
-            body = 'AnkiConnect v.{}'.format(util.setting('apiVersion')).encode('utf-8')
+            try:
+                filename = list(req.headers)[0].decode('utf-8').split()[1]
+                if filename == "/":
+                    body = 'AnkiConnect v.{}'.format(util.setting('apiVersion')).encode('utf-8')
+                else:
+                    print(filename)
+                    params = {
+                            "action": "retrieveMediaFile",
+                            "params": {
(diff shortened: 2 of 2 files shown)


FooSoft:
Hi there! What kinds of problems are you running into, exactly? You should just be able to execute the `retrieveMediaFile` action (as you are doing in the PR yourself), and get the data out right? You should be able to just make the request using JavaScript embedded in your web page. If there are usability problems, we should address them on the `retrieveMediaFile` level. The stuff in `web.py` is really only meant to be a minimal framework for handling Anki API calls; I would be very hesitant to add effectively a secondary channel for using the API.

corollari:
The main problem is that the implementation used to get the media objects present in a card is really complex (in my opinion it would at the very least require an html parser and a js runtime to catch all edge cases) but it could be vastly simplified if this PR were to be merged.

FooSoft:
Are you talking about the case where you don't know what media is present on a card, but want to just download everything? I'm not sure I follow. If you already know the filename that you want to get (which your PR assumes you do), at what point do you have to do HTML parsing?

corollari:
I see how what I am saying can be confusing, let me rephrase it. My end goal is to display cards outside of anki, concretely in a browser, and to do that I need to download all the media inside these cards. Here's the different ways to do that:
- **If this PR gets merged**: I can just do `body.innerHTML=cardHTML` and the browser will take care of fetching all the resources used in the card by sending normal HTTP requests
- **Without this PR**: I need to do all these things:
  - Parse the card's html, request the resources and put the in the html
  - Parse the css to look for fonts specified in urls, request them and replace them in the css file
  - Make sure that javascript code doesn't change something that will cause another file to be requested. If it does, I need to catch that and replace these resources as well.
  - I'm probably still missing edge cases that will then lead to some cards breaking

Does this clarify the reasoning behind this PR?

FooSoft:
Yes, I see what you are trying to do now. I need to think about this some more. I don't like the idea of AnkiConnect extension becoming a general purpose web server for Anki content. That seems like it is opening a can of worms.

Kubo2:
@corollari I am tackling the same issue. I have found it is possible to intercept, change, and redirect HTTP requests made by the browser in JavaScript using a [ServiceWorker](https://stackoverflow.com/a/43815800/3827190) (or even simpler, the [webRequest API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Intercept_HTTP_requests)). After intercepting a request, you can redirect it back to your server that will retrieve any necessary media files from Anki.

corollari:
Thanks, I'll look into it. Last time I search for something that would let me intercept requests I wasn't able to find something that would work in chrome but this looks promising.

Issue #117, opened 2019-05-07
Add stats resource to api


Fandekasp:
Hi,

I found AnkiConnect while looking for an API to query stuff on Anki. The API you offer is intuitive, and I'd really appreciate if we could also use it to query statistics (eg how many cards have I studied in how many minutes today, how many days have I studied this month etc).

Do you think this is something that could fit with anki-connect? Or do you know a better way to query this information from the shell?

FooSoft:
It's definitely something that could fit into the API. There isn't really isn't another easy way of doing it, so if you'd like to take a crack at it, feel free to send a PR and I'll merge it in 👍

cjdduarte:
I was going to request the same, but I found this post here.

cjdduarte:
> It's definitely something that could fit into the API. There isn't really isn't another easy way of doing it, so if you'd like to take a crack at it, feel free to send a PR and I'll merge it in 👍

I will evaluate here and post in the future

manuelsinn:
> 
> 
> > It's definitely something that could fit into the API. There isn't really isn't another easy way of doing it, so if you'd like to take a crack at it, feel free to send a PR and I'll merge it in 👍
> 
> I will evaluate here and post in the future

Hey there, I am trying right now to get stats with anki connect.
First of all, great work @FooSoft, you have really made a lot of people's lives easier!
Some things have been pretty easy to implement, like simple queries to get today's reps or the number of mature cards (useful because for me that is basically how many words I really know).

Something I am struggling with is to find out the current streak: to calculate this precisely I basically have to send a request for each day of the streak, which is obviously really inefficient.

@AdrienLemaire Maybe we can work something out together, to make queries of a set of useful statistics available?

@FooSoft Something that seems like an issue to me is that the client needs to run in order to gather the data. To showcase the stats on a web page and have them updated frequently it would make sense to have an option to send a request on start-up, or whenever anki is syncing. Would something like that be possible?

pavelgordon:
Hey @manuelsinn, I am looking exactly at this functionality right now: exporting stats somehow. 
I would be satisfied with just having numbers of mature cards. Any chance that you could share what you have done already?

manuelsinn:
Hey @pavelgordon, I just made the script I created available in this repo: 
https://github.com/manuelsinn/anki-stats
Hope it helps! Would be excited to create a more usable version together (especially with the issue I mentioned above).

PR #155, opened 2020-04-19
add api to get number of cards reviewed in the current day

yakrider:
partly addresses issue raised in #117
(being able to get number of cards reviewed in the current day allows one to use AnkiConnect for automation for things like blocking websites etc until a certain number of cards have been reviewed everyday etc)

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -285,6 +285,11 @@ class AnkiConnect:
         return list(map(self.handler, actions))
 
 
+    @util.api()
+    def getNumCardsReviewedToday(self):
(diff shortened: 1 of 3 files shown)


plugin/__init__.py
```diff
@@ -285,6 +285,12 @@ def multi(self, actions):
         return list(map(self.handler, actions))
 
 
+    @util.api()
+    def cardsReviewedToday(self):
```
FooSoft, line 289: Please rename to `getCardsReviewedToday` to be consistent with *most* APIs that start with a verb.

FooSoft:
Thanks!

Issue #109, opened 2019-02-26
endpoint: Get profile name


bikenik:
Hello. Could you add the endpoint to return current profile name? I've got several profiles and I toggle them all the time. I need it to dynamic change the path in my app to `collection.media` folder. Hope it should be useful for others also.
for example: 

```py
@api()
def loadProfileName(self):
    return self.window().pm.name
```

4a57b2c482b889ec8750817d836d939bc82059fa

or return Array of existing Profile names instead `True` or `False`  is even better.

```py
    @api()
    def loadProfile(self, name):
        if name not in self.window().pm.profiles():
            return False
        if not self.window().isVisible():
            self.window().pm.load(name)
            self.window().loadProfile()
            self.window().profileDiag.closeWithoutQuitting()
        else:
            cur_profile = self.window().pm.name
            if cur_profile != name:
                self.window().unloadProfileAndShowProfileManager()
                self.loadProfile(name)
        return self.window().pm.profiles()
```

bd1b5a0c6ffe40d8ef05e17a1f0be247a79c6bce

Anyway, I face off with a case when I need to get/pick out profile name/names. And I can't do it with current API endpoints.
Thnx in advance!

PR #159, opened 2020-05-01
Add getProfiles action

snoyes:
Returns the list of profiles, to complement the existing loadProfile action.


snoyes:
Would also address #109 which was closed without response for some reason.

PR #160, opened 2020-05-01
Import package

snoyes:
Add an importPackage to complement exportPackage.

The motivation for this and the getProfiles PR #159 are to support cases like a teacher who has access to all their student's profiles, and wishes to automate looping through them all to import a deck or to sync the Leaderboard add-on.

https://www.reddit.com/r/Anki/comments/ga6zdp/using_anki_in_schools_or_bulding_an_alternative/

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,30 @@ guarantee that your application continues to function properly in the future.
     }
     ```
 
+*   **importPackage**
+
+    Imports a file in `.apkg` format into the collection. Returns `true` if successful or `false` otherwise.
+    Note that the file path is relative to Anki's collection.media folder, not to the client.
+
+    *Sample request*:
+    ```json
+    {
+        "action": "importPackage",
+        "version": 6,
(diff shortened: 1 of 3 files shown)


FooSoft:
Both this and #159 look good -- could you also hook up a simple unit test so that these APIs are covered?

snoyes:
Unit test for #159 should be no difficulty.

For this one, is it acceptable to load the deck created by the export test? Or do you want all the tests to be independent of each other?

FooSoft:
Yup, that should be fine! The main goal is to be able to catch AnkiConnect breakage from code churn in Anki (class members being moved/renamed/deleted).

snoyes:
Turned out to be easier to just create a deck for the test, since the existing exportPackage test uses the Default deck, and it's not trivial to check that it actually imported anything when it skips duplicates.

FooSoft:
Awesome, thanks!

PR #166, opened 2020-05-03
Collection stats

snoyes:
Return the collection statistics. Users may then parse that report for all the details they want about cards reviewed and times and whatnot instead of asking for an end point for each one.

It's unfortunate that Anki wraps the HTML presentation of the report and the queries to fetch the data into a single method, but I think it's preferable to let the users worry about running the result through BeautifulSoup or something instead of trying to maintain the queries to get the data out in a more consumable format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,14 +370,14 @@ guarantee that your application continues to function properly in the future.
     }
     ```
 
-*   **getCollectionStats**
+*   **getCollectionStatsHTML**
 
     Gets the collection statistics report
 
     *Sample request*:
     ```json
     {
-        "action": "getCollectionStats",
+        "action": "getCollectionStatsHTML",
         "version": 6
     }
     ```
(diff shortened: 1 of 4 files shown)


FooSoft:
Yup -- agreed. That being said, maybe `getCollectionStatsReport` or `getCollectionStatsHtml` would be a better name (to emphasize that the caller is just getting processed data which requires parsing to use)?

snoyes:
I'm inclined to go with ...Html, because then if Anki separates the layers and makes the base data available, we can adjust this one to maintain the API as is and add another to expose that base data.

FooSoft:
Makes sense. The data is already all there in Anki, but it's just a question of how deep you want to dig. I really wish Anki provided a way to do all these things without just poking at random data and functions in the process (which inevitably break due to code churn...)

PR #169, opened 2020-05-08
collection stats - whole collection

snoyes:
By default, the collection stats report retrieves only stats for the most recently viewed deck. Add a parameter to fetch stats for the whole collection.


FooSoft:
Looks good! :+1:

PR #171, opened 2020-05-08
Do not strip() during dupe check

### LINKED ISSUE #164: Can't prevent duplicates from being added via API
First - THANKS so much for creating this API, it's been a lifesaver for me as I try to turn my kindle highlights into anki cards. 

One issue I'm having is the ability to prevent duplicate cards from being created. I've set the flag to 'False' but it still allows me to create a duplicate cards. 

Here's an example of the JSON object I'm passing through -> https://github.com/matthewwoo/kindle_anki/blob/master/models.py

snoyes:
Fix #164 

The deck-level duplicate check was stripping whitespace before comparing the value. Anki does not do that. Anki only checks if the field would be empty if the whitespace were to be stripped; duplicate comparison is done with whitespace in place.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -236,7 +236,7 @@ class AnkiConnect:
 
         # dupeOrEmpty returns if a note is a global duplicate
         # the rest of the function checks to see if the note is a duplicate in the deck
-        val = note.fields[0].strip()
+        val = note.fields[0]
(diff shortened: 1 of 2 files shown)


KngZhi:
Thank you so much. This issue caused some problems before, Now I do not need to truncate duplicate notes anymore.

PR #207, opened 2020-11-23
Add support for video and pictures when adding or updating a note

grepmew:
Sound and video share same [sound:..] tag. See https://github.com/ankitects/anki/blob/0c08ff53179ac1d242b845457db4e4414aef2662/pylib/anki/sound.py#L33


FooSoft:
Awesome, thanks!

PR #208, opened 2020-11-25
Add file parameter to storeMediaFile

an0:
Make it easier to add local media files.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -468,10 +468,15 @@ class AnkiConnect:
 
 
     @util.api()
-    def storeMediaFile(self, filename, data=None, url=None):
+    def storeMediaFile(self, filename, data=None, path=None, url=None):
         if data:
             self.deleteMediaFile(filename)
             self.media().writeData(filename, base64.b64decode(data))
+        elif file:
+            self.deleteMediaFile(filename)
+            with open(path, 'rb') as f:
+                data = f.read()
+            self.media().writeData(filename, data)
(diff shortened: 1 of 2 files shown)


plugin/__init__.py
```diff
@@ -468,10 +468,15 @@ def removeDeckConfigId(self, configId):
...
 
     @util.api()
-    def storeMediaFile(self, filename, data=None, url=None):
+    def storeMediaFile(self, filename, data=None, file=None, url=None):
         if data:
             self.deleteMediaFile(filename)
             self.media().writeData(filename, base64.b64decode(data))
+        elif file:
+            self.deleteMediaFile(filename)
+            f = open(file, 'rb')
```
FooSoft, line 477: Please change to something like:
```
with open(file, 'rb') as fp:
    self.media().writeData(filename, f.read())
```
This will ensure that the file is properly closed.

plugin/__init__.py
```diff
@@ -468,10 +468,15 @@ def removeDeckConfigId(self, configId):
 
 
     @util.api()
-    def storeMediaFile(self, filename, data=None, url=None):
+    def storeMediaFile(self, filename, data=None, file=None, url=None):
```
FooSoft, line 471: Looking at this a little bit further, we probably need to have a better name for `file`, as it does not make it clear what it is being used for. I would recommend `dataPath` to make it clear that it is where the data is coming from. The docs will need to also be updated to reflect this change.

Issue #209, opened 2020-11-26
Server stops responding if client disconnects abruptly


kanjieater:
By starting requests and then cancelling them Anki Connect will eventually stop responding.

Then while anki connect is frozen, after about 5-10mins I got this error from ankiconnect in anki:
`ConnectionResetError: [WinError 10054] An Existing connection was forcibly closed by the host`

That error popup then crashed anki and I couldn't even copy the text out of the error box, and had to copy it by hand while it was frozen - trying to get more detailed crash log, but I'm not sure how if the gui is crashing.

I'm doing basic cardInfo queries over localhost:

```
{
    "action": "cardsInfo",
    "version": 6,
    "params": {
        "cards": [
            1198884699985,
            1198899930593
        ]
    }
}
```

If I do multiple of these from the client I'm trying to use (GoldenDict) or open up a couple tabs with postman I can get this to happen. Start a query like the one above then cancel it and start another, eventually ankiconnect and anki freeze. It seems like anki stays working even while ankiconnect is frozen at first, but then both crash. It's easier to reproduce on large queries, so adding more card id's to the above can help reproduce.

```
    
    Anki 2.1.22 (0ecc189a) Python 3.8.0 Qt 5.14.1 PyQt 5.14.1
    Platform: Windows 10
    Flags: frz=True ao=True sv=1
    Add-ons, last update check: 2020-11-25 18:28:09
    
    
    ===Add-ons (active)===
    (add-on provided name [Add-on folder, installed at, version, is config changed])
    '' ['PassFailJP', 0, 'None', '']
    '' ['ResetEZ', 0, 'None', '']
    '' ['anki_image', 0, 'None', '']
    '' ['autoLapseNewInterval', 0, 'None', '']
    '' ['autobulk', 0, 'None', '']
    '' ['autovocab', 0, 'None', '']
    '' ['bulkclearhtml', 0, 'None', '']
    '' ['jsentencehighlighter', 0, 'None', '']
    '' ['kanjivocab', 0, 'None', '']
    '' ['leech_handler', 0, 'None', '']
    '' ['phonetics', 0, 'None', '']
    '' ['rtk', 0, 'None', '']
    Advanced Browser ['874215009', 2020-03-14T16:25, 'None', '']
    Advanced Copy Fields ['1898445115', 2019-05-05T13:25, 'None', '']
    Anki Simulator ['817108664', 2020-07-21T09:17, 'None', '']
[...]

PR #217, opened 2020-12-26
Server no longer hangs on client disconnects

kanjieater:
As mentioned here https://github.com/FooSoft/anki-connect/issues/209 there's an easy to reproduce critical error that causes the server to hang if the client disconnects. I've been using the PR in this code for about a month and it continues to work without crashing or ill side-effects.

### DIFF
--- a/plugin/web.py
+++ b/plugin/web.py
@@ -46,13 +46,13 @@ class WebClient:
             return False
 
         rlist, wlist = select.select([self.sock], [self.sock], [], 0)[:2]
-        self.sock.settimeout(1.0)
+        self.sock.settimeout(5.0)
 
         if rlist:
             while True:
                 try:
                     msg = self.sock.recv(recvSize)
-                except: #ConnectionResetError or socket.timeout
+                except (ConnectionResetError, socket.timeout):
                     self.close()
                     return False
                 if not msg:
--- a/plugin/web.py
+++ b/plugin/web.py
@@ -19,7 +19,6 @@ import socket
 
 from . import web, util
 
-
 #
 # WebRequest
 #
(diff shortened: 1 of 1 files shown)


plugin/web.py
```diff
@@ -47,27 +46,38 @@ def advance(self, recvSize=1024):
...
 
         if rlist:
-            msg = self.sock.recv(recvSize)
-            if not msg:
-                self.close()
-                return False
+            while True:
+                try:
+                    msg = self.sock.recv(recvSize)
+                except: #ConnectionResetError or socket.timeout
```
FooSoft, line 55: Please explicitly catch these exceptions and let other ones go through. Don't want to do a catch-all here and doing so is stronger than having a comment.

plugin/web.py
```diff
@@ -47,27 +46,38 @@ def advance(self, recvSize=1024):
             return False
 
         rlist, wlist = select.select([self.sock], [self.sock], [], 0)[:2]
+        self.sock.settimeout(1.0)
```
FooSoft, line 49: This seems to be a bit aggressive; I agree that this will work for most situations (especially considering that in 99.9% of the cases we will be connected to localhost), but it might still be good to bump this up to 5 seconds or so.

FooSoft (changes requested):
Looks good overall, just a couple of nitpicks / thoughts

plugin/web.py
kanjieater: Done!

plugin/web.py
kanjieater: I thought this slowed things down previously, but upon testing just now, it does seem like everything works without lag when increasing the timeout.

kanjieater (commented):
Fair points. I've updated and tested, and it still works as expected. Thank you.

PR #228, opened 2021-02-03
Added findAndReplaceInModels

itsmepvr:
Hey @FooSoft ,
### Added `findAndReplaceInModels` api call.
* To find and replace string in existing models by name or all models by name - 'all_models'.
* Also added needed documentation in models.
* Referred and Implemented this [addon](https://ankiweb.net/shared/info/992153575).


plugin/__init__.py
```diff
@@ -918,6 +918,41 @@ def updateModelStyling(self, model):
         models.flush()
 
 
+    @util.api()
+    def findAndReplaceInModels(self, modelName, findText, replaceText, Front = True, Back = True, css = True):
+        if modelName == 'all_models':
```
FooSoft, line 923: Just make `modelName` be optional; if not provided, it means all models.

plugin/__init__.py
```diff
@@ -918,6 +918,41 @@ def updateModelStyling(self, model):
...
+                        if Back and key == 'afmt' and findText in value:
+                            checkForText = True
+                            tmpls[key] = value.replace(findText, replaceText)                       
+            self.collection().models.save(model, True)
+            self.collection().models.flush()
+            if checkForText:
+                updatedModels += 1
+        if updatedModels == len(ankiModel):
+            return "Found and replaced in {0} models".format(updatedModels)
+        else:
```
FooSoft, line 952: Don't need `else`

plugin/__init__.py
```diff
@@ -918,6 +918,41 @@ def updateModelStyling(self, model):
         models.flush()
 
 
+    @util.api()
+    def findAndReplaceInModels(self, modelName, findText, replaceText, Front = True, Back = True, css = True):
```
FooSoft, line 922: Don't do uppercase for `Front` and `Back`, no spaces around `=` in argument lists.

plugin/__init__.py
```diff
@@ -918,6 +918,41 @@ def updateModelStyling(self, model):
...
+        for model in ankiModel:
+            model = self.collection().models.byName(modelName)
+            checkForText = False
+            if css and findText in model['css']:
+                checkForText = True
+                model['css'] = model['css'].replace(findText, replaceText) 
+            for tmpls in model.get('tmpls'):
+                for key, value in tmpls.items():
+                    if key in ('qfmt', 'afmt'):
+                        if Front and key == 'qfmt' and findText in value:
```
FooSoft, line 940: Do explicit checks `is None` or `is not None` for `front` and `back`.

plugin/__init__.py
```diff
@@ -918,6 +918,41 @@ def updateModelStyling(self, model):
...
+                            tmpls[key] = value.replace(findText, replaceText)
+                        if Back and key == 'afmt' and findText in value:
+                            checkForText = True
+                            tmpls[key] = value.replace(findText, replaceText)                       
+            self.collection().models.save(model, True)
+            self.collection().models.flush()
+            if checkForText:
+                updatedModels += 1
+        if updatedModels == len(ankiModel):
+            return "Found and replaced in {0} models".format(updatedModels)
```
FooSoft, line 951: Why return a formatted string? That can be annoying for caller that may want to use the value. Just return the count.

FooSoft (changes requested):
Also please add unit test.

tests/test_models.py
```diff
@@ -57,4 +57,4 @@ def runTest(self):
 
 
 if __name__ == '__main__':
-    unittest.main()
+    unittest.main()
```
FooSoft, line 60: Was there meant to be a test added in this file?

itsmepvr: Yes, still need to do it. Actually I never wrote unit tests before. Will do it.

itsmepvr: @FooSoftm Hey, I added unit test, let me know if that is correct cause I am a newbie writing unit tests.

itsmepvr: Added unit tests. Let me know if it is correct cause I am a newbie for writing unit tests. Thanks

plugin/__init__.py
FooSoft: No I mean just make it `None`, this way if someone has a `modelName` of `all_models` it won't be an issue. There's no reason to make this be an actual special string. Also, I see that `findText` and `replaceText` default to empty string, why? I think if you are calling `findAndReplaceInModels` it's reasonable to assume that `findText` and `replaceText` are provided.

tests/test_models.py
FooSoft: Tests themselves look fine I think. 👍🏻

FooSoft (changes requested):

plugin/__init__.py
itsmepvr: Updated..!! Please check

PR #229, opened 2021-02-07
Fix bug in loadProfile method

lazy-void:
When sync is enabled in profile, `loadProfile` fails. It happens because main window is still active during sync. The `loadProfile` method recursively calls itself over and over and tries to logout from profile each time (calling method                   `self.window().unloadProfileAndShowProfileManager()`). These actions overload Anki app.

My solution is to use a timer with an interval of 1 second to check if the main window is still active.

bug_loadProfile


FooSoft:
Looks good, thanks!

PR #230, opened 2021-02-13
Allow to add media to notes by file path or base64 encoded data

davidsauerwein:
The existing `storeMediaFile` action allows adding media using the `data`, `path` or `url` parameter. The current `addNote` implementation only allows a `url` parameter, however.

This pull request uses the existing code for `storeMediaFile` to add media when using `addNote` and therefore also allows using the `data` and `path` parameters in addition to `url`.

In case you want to merge this I will of course also adjust the documentation :smiley:

### DIFF
--- a/actions/media.md
+++ b/actions/media.md
@@ -20,7 +20,7 @@
     *Sample result*:
     ```json
     {
-        "result": null,
+        "result": "_hello.txt",
         "error": null
     }
     ```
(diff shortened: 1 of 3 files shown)


FooSoft:
I think this is a good change, if you update the docs I'll merge it in 👍🏻

davidsauerwein:
> I think this is a good change, if you update the docs I'll merge it in 👍🏻

Done

PR #234, opened 2021-02-25
Add "forgetCards" and "relearnCards"

hrdrq:
## Use case of the new features
Sometimes there are some words I don't understand and after trying adding them to the deck,
I found out that they were already in the deck and I had studied them.
In the case, I want to study them again ASAP, so resetting the cards or making them "relearning" is convenient.

## About test codes
I tried to write the unit test, but there is no easy way to change the card's state.
If there is some method to change the card to "review", it'll be useful.
(Codes in `tests/test_cards.py` is only creating "new" cards)
Do you have any ideas?
Thank you

### DIFF
--- a/actions/cards.md
+++ b/actions/cards.md
@@ -306,3 +306,49 @@
         "error": null
     }
     ```
+
+*   **forgetCards**
+
+    Forget cards, making the cards new again.
+
+    *Sample request*:
+    ```json
+    {
+        "action": "forgetCards",
+        "version": 6,
+        "params": {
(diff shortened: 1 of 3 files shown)


FooSoft:
If you could just create some simple tests that call the new APIs that would be sufficient, I think. The main thing that I am interested in getting out of the unit tests is detecting when the plugin stops working because of Anki changes.

hrdrq:
Thank you! I had added the tests.

FooSoft:
Looks good, thanks!

PR #235, opened 2021-02-25
Fixed bug with skipHash causing media not to be added to notes

ripose-jp:
When `skipHash` is used with an action like `addNote`, hashlib would throw `object supporting the buffer API required`.
This happened because the `data` variable was accidentally being used to calculate hashes instead of `mediaData`.


FooSoft:
Thanks!

PR #236, opened 2021-02-26
Use proper variable when computing skipHash

Rapptz:
If data is not passed but a URL is, the skipHash check will still use
the data variable instead of the one containing actual data (mediaData).

This fixes that issue.


Rapptz:
Just noticed this is a duplicate of #235. Closing. Apologies.

PR #238, opened 2021-02-28
Fix handling of errors in several methods

lazy-void:
If you try to get a note from Anki by an invalid `id`, it will raise `NotFoundError` with an empty message. The current implementation assumes you are getting `None`, and therefore `handler` catches `NotFoundError` instead of our custom `Exception` with a meaningful message.

### DIFF
--- a/actions/cards.md
+++ b/actions/cards.md
@@ -94,18 +94,18 @@
         "error": null
     }
     ```
+    
+*   **suspended**
 
-*   **areSuspended**
-
-    Returns an array indicating whether each of the given cards is suspended (in the same order).
-
+    Check if card is suspended by its ID. Returns `true` if suspended, `false` otherwise.
+    
     *Sample request*:
     ```json
     {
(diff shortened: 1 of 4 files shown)


lazy-void:
I found same bugs in `cardsInfo` and `notesInfo` methods and fixed them.

lazy-void:
Hm... It seems like a lot of methods which call `self.collection().getCard()` or `self.collection().getNote()` methods have these problems because they make wrong assumptions.
I will create wrappers for these methods in the AnkiConnect class. They will either return a result or raise an exception with a descriptive message. This should simplify further maintenance.

FooSoft:
Nice work, thanks!

thovip119:
I have this problem after the update, it's gone when I update Anki, but my other add-ons required lower anki version. 
image

FooSoft:
@thovip119 yup, I see the problem. Just pushed an update that should fix things. Let me know if you are still having any issues.

PR #239, opened 2021-02-28
added getNumCardsReviewedByDay action

qharscoet:
Might add options to filter by decks or something but for now seems to be usable


actions/statistics.md
```diff
@@ -20,6 +20,29 @@
...
+        "action": "getNumCardsReviewedByDay",
+        "version": 6,
+    }
+    ```
+
+    *Sample result*:
+    ```json
+    {
+        "result":  [
+            ['2021-02-28', 124],
```
FooSoft, line 39: These are single quotes, so it's not valid JSON

plugin/__init__.py
```diff
@@ -337,6 +337,11 @@ def multi(self, actions):
     def getNumCardsReviewedToday(self):
         return self.database().scalar('select count() from revlog where id > ?', (self.scheduler().dayCutoff - 86400) * 1000)
 
+    @util.api()
+    def getNumCardsReviewedByDay(self):
+        return self.database().all("select date(id/1000 - ?, 'unixepoch', 'localtime') as day, count() from revlog group by day order by day desc",
```
FooSoft, line 342: Single quotes in python strings please

FooSoft (changes requested):

FooSoft:
Please also add some basic tests to make sure that we are at least calling these APIs.

qharscoet:
Indeed, sorry for the wrong quotes.
Also added basic test inspired by the ones for the other stats actions.

FooSoft:
Looks good, thanks!

PR #242, opened 2021-03-13
Compare field names in a case-insensitive manner

abdnh:
Field names should be compared in a case-insensitive manner, or if we were pedantic, with full case folding, but even Anki doesn't bother doing that in all cases (except maybe when creating fields), so lowercasing should be fine.


FooSoft:
Looks good!

PR #244, opened 2021-03-29
Dev

itsmepvr:
Hey @FooSoft,
Updated replaceTags function by combining both replaceTags and replaceTagsInAllNotes. Please check...:)

### DIFF
--- a/actions/notes.md
+++ b/actions/notes.md
@@ -330,6 +330,30 @@
     }
     ```
 
+*   **replaceTagsInAllNotes**
+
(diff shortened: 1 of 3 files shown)


FooSoft:
How will this impact extensions that are currently using `replaceTagsInAllNotes`? Won't this change break them as soon as the user updates AnkiConnect?

itsmepvr:
That's a valid point..!!
I added back "replaceTagsInAllNotes" and redirected that to "replaceTags" so this update will not impact any extensions that are currently using.

PR #247, opened 2021-04-09
Added "getMediaFilesNames" action

truthseeker-skv:
@FooSoft
Added an action to get a list of the names of media files matching the passed pattern. Returning all names by default.

### DIFF
--- a/actions/media.md
+++ b/actions/media.md
@@ -93,6 +93,29 @@
     }
     ```
 
+*   **getMediaFilesNames**
+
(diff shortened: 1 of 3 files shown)


lazy-void:
I think it would be much better if pattern matching was done using regular expressions.

FooSoft:
I think it's fine -- wildcards are what you normally have to work with when doing file operations. Thanks for the submission 👍🏻

PR #250, opened 2021-04-25
Update createModel method to support Cloze models

lazy-void:
I've updated the `createModel` method so that it can be used to create Cloze note types.


lazy-void:
Can you check this out, @FooSoft?

FooSoft:
Looks good to me, sorry for delay!

PR #252, opened 2021-05-04
Fix vulnerability

DegrangeM:
The current code to handle CORS is defective.

**It is currently possible to execute action even if origin is not allowed.** The browser will refuse to give the response to javascript (but it will still travel on the network), but the requested action will **still be executed**. So **it's possible to delete deck**, upload big or malicious file, etc.

This PR fix that by not executing the request action and returning a 403 error if the origin is not allowed.

Test effectued:
- [X] Api refuse the request if called from a not allowed origin (javascript)
- [X] Api handle the request if called from an allowed origin (javascript)
- [X] Api handle the request (python does not set origin)
- [X] Api handle the request if called from 127.0.0.1 and localhost allowed (javascript)
- [X] Api handle the request if called from google chrome or firefox extension and localhost allowed (javascript)

I added support for 127.0.0.1 and browser extension for compatibility reason, but for best security it shouldn't.

-----

This PR also include an improvement and simplification of the default code (https://github.com/FooSoft/anki-connect/commit/2a0010f0730d4e0ebe4af1609bd89fa378e0d518):
→ if `*` is in the list of allowed origin, set `Access-Control-Allow-Origin` to `*` even if the origin of the request is also in the list
→ if origin is not set in the request, return `http://localhost` even if there is only one element in the allowed origin list
The code was strange but I am not an expert so maybe there was a reason to these cases. Feel free to revert this commit.

### DIFF
--- a/plugin/config.json
+++ b/plugin/config.json
@@ -3,6 +3,5 @@
     "apiLogPath": null,
     "webBindAddress": "127.0.0.1",
     "webBindPort": 8765,
-    "webCorsOrigin": "http://localhost",
     "webCorsOriginList": ["http://localhost"]
 }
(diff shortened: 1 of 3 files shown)


FooSoft:
Thanks for this high quality contribution 💯

PR #255, opened 2021-05-07
Add requestPermission API method

### LINKED ISSUE #254: Allow website to request to be added to trusted origin
Hi,

Currently, editing the configuration to add a trusted origin is not user friendly.

I suggest to add a method `requestPermission` that could be called to request to be added to trusted origin.

**I am willing to add the functionnality**, I am just asking before what you think and if you have any comment.

When this method is called, a message dialog would appear (and be focused). It will display something like "http://example.com want to request permission to use AnkiConnect. Do you want to give it access ? By giving permission, the website will be able to do action on anki, including deleting your deck." (if you have a better message feel free). There would be a yes button and a no button (and probably a block button ?)

If pressing yes, the origin will be added to the list of trusted origin.

The function will return 

```json
 {
    "permission" : "granted/denied",
    "requireApikey" : "true/false"
}
```

if already trusted it would return 

```json
 {
    "permission" : "granted",
    "requireApikey" : "true/false"
}
```

This function will of course, don't be affected by CORS.

DegrangeM:
This allow website to request permission to use the Api by displaying a popup to the user.

image

- [x] Calling the method from an already trusted origin work
- [x] Calling the method from an untrusted origin work
- [x] Calling the method from an ignored origin doesn't display the popup
- [x] Calling an other api from an already trusted origin work  

Fix #254

### DIFF
--- a/actions/miscellaneous.md
+++ b/actions/miscellaneous.md
@@ -44,10 +44,6 @@
 
 Gets the version of the API exposed by this plugin. Currently versions `1` through `6` are defined.
 
-This should be the second call you make to make sure that your application and AnkiConnect are able to communicate
-properly with each other. New versions of AnkiConnect are backwards compatible; as long as you are using actions
-which are available in the reported AnkiConnect version or earlier, everything should work fine.
-
 *Sample request*:
 ```json
 {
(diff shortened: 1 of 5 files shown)


FooSoft:
Looks good!

PR #257, opened 2021-05-09
storeMediaFile: Add option to keep file with the same name

abdnh:
This adds the `deleteExisting` field to the storeMediaFile action to optionally keep a file with the same name in the media folder intact, relying on Anki to rename the added file to make it unique.


FooSoft:
Looks good!

PR #262, opened 2021-06-05
Add `updateNotesFields`

archiif:
Does the same thing as `updateNoteFields` but for multiple notes.

### DIFF
--- a/actions/notes.md
+++ b/actions/notes.md
@@ -217,6 +217,46 @@
     }
     ```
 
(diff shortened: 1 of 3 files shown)


FooSoft:
You should already be able to do this with the `multi` action. I don't really want to add specific plural actions, since this would set a precedent for more similar things being added to the API for all other actions.

archiif:
Ah, didn't know that was a thing.

Issue #275, opened 2021-07-19
guiAddCards vs. addNote


mzhou08:
As a new user of Anki, I was trying to use ankiConnect to automatically add cards to my decks without opening any dialogs. I found the guiAddCards function, and tried to modify it to not show dialogs. However, I found the addNote function, which accomplishes exactly what I wanted.

To prevent similar confusion for future users, I've slightly modified the guiAddCards command to include an autoAdd option, which directly calls addNote. I've updated the README accordingly, adding a few lines to guiAddCards's description. I've also enabled silent closing for the closeAfterAdding.

Thanks for developing such a comprehensive Anki API!

mzhou08:
Please review my pull request, linked [here.](https://github.com/FooSoft/anki-connect/pull/276)

uzqw:
yes, I have searched for the function of adding cards without GUI for so long, until I find your PR. "Add notes" really confused me, I didn't know it was what I need.

PR #276, opened 2021-07-19
Pull request for Issue #275

mzhou08:
To prevent confusion for users between guiAddCards and addNote, I've slightly modified the guiAddCards command to include an autoAdd option, which directly calls addNote. I've updated the README accordingly, adding a few lines to guiAddCards's description. I've also enabled silent closing for the closeAfterAdding.


introt (changes requested):
This issue concerns those who are not familiar with the distinction between notes and cards; thus the solution is user education, not more code.

Please remove the code additions which introduce unnecessary maintenance burden for the developers, and find a way to introduce the [key concepts](https://docs.ankiweb.net/getting-started.html#key-concepts) (or a link to such introduction) to the new user instead. Thanks for your consideration.

PR #281, opened 2021-08-24
Update README.md

Twinov:
The searching documentation seems to have moved from `https://docs.ankiweb.net/#/searching` to `https://docs.ankiweb.net/searching.html`.


FooSoft:
Thanks!

PR #282, opened 2021-08-25
feat: cardsInfo now also outputs the fields without html

thiswillbeyourgithub:
- feat: cardsInfo now also returns the modification time
- minor: wrong indentation
- new method: get modification time of cards
- docs: mentions new method cardsModTime
- docs: readme now mentions that modification time is in the result from cardsInfo
- feat: cardsInfo now also returns the fields with stripped media
- style: rename no_media to strip_media

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,36 @@ corresponding to when the API was available for use.
     }
     ```
 
+*   **cardsModTime**
+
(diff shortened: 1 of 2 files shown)


thiswillbeyourgithub:
I'm sorry I seem to have wrongly done my PR, my apologies :( !

Issue #122, opened 2019-09-06
Any way to change the position of a card in the new cards queue?


adaszko:
Hi and thank you for creating AnkiConnect.

I couldn't find any way to change a position of a card in the queue which is something I'd find very useful given my deck has thousands of unseen cards.  Is there a way to do that?  Are there plans to implement it if it's not currently possible?

Thank you!

sgallag-insta:
I also was struggling to find an answer to this question so I will post an answer here in case anyone else stumbles across this:

If you just want to move a card to the beginning of the queue, you can use `relearnCards`.
If you want more fine-grained control of the queue, it looks like you should be able to use `updateCompleteDeck`, although I haven't tried this myself.

TobiasWehrum:
For what it's worth, I've since started using #283's `setSpecificValueOfCard` for exactly this and so far it seems to be working well. Thanks for making it, @thiswillbeyourgithub!

PR #283, opened 2021-08-25
misc + new methods: cardsInfo now also returns the fields with stripped media + createFilteredDeck + runConsoleCmd + setSpecificValueOfCard + bury / unbury 

thiswillbeyourgithub:


thiswillbeyourgithub:
Regarding the latest commit:e4f893b I was thinking of maybe, if "always" is set, generating a random number and if greater than a threshold then warn the user anyway. This would avoid having user letting the option on and forgetting about it. Let me know if you want me to add this.

thiswillbeyourgithub:
Hi, [the project I created](https://github.com/thiswillbeyourgithub/AnnA_Anki_neuronal_Appendix) that needed some features from Anki-Connect is now finished. It would be great if you could review my changes so that I can begin to publicly talk about it.

Currently I am the only on to be able to use my script because my modification to anki-connect is not yet centralized here :)

Let me know if I can be of any help!

thiswillbeyourgithub:
up :)

thiswillbeyourgithub:
Hi, with no answer from you I'm more and more thinking about forking anki connect to make a distinct anki addon for my specific use.

Is that okay for you? (I would of course be very clear that it's just a minimalized fork and pay credits when due :) )

FooSoft:
You are certainly more than welcome to fork. As it stands, the PR is pretty large and there are many issues with it. Unfortunately going over the design and naming of individual APIs, checking coding style, and validating that everything works takes a solid block of time which I never seem to have. This is why this has been getting punted for a while.

Some top level issues:
* No unit tests
* Style inconsistency even within this change alone (for example single quotes vs double quotes)
* Weird parameterization of actions like `createFilteredDeck` and `runConsoleCmd` (what good does the token do)
* Still not happy with security implications of `runConsoleCmd`... you can't even see what command is about to be executed.

If you break this up into smaller bits we can go over them individually.

TobiasWehrum:
@thiswillbeyourgithub, do you think it might be possible to resubmit `setSpecificValueOfCard`? It's the only way I found to set the due date of a card (see also https://github.com/FooSoft/anki-connect/issues/122).

thiswillbeyourgithub:
done

PR #287, opened 2021-09-07
Save note after field update

marcel-cd:
I use anki-connect from Emacs (anki-editor). Often the command updateNoteFields did not persist the changes. Perhaps this fix help.


FooSoft:
Does it help?

marcel-cd:
Yes, since I made the change, all updates work as expected.

PR #288, opened 2021-09-14
guiBrowse: Fixes MockModel error on new Anki versions

### LINKED ISSUE #277: guiBrowse returns error on Anki versions past 2.1.45
I'm currently trying out the latest Anki (2.1.45+), and I noticed that the guiBrowse action in anki-connect returns an error now, regardless of what query I send it.

Example request
```
{
    "action": "guiBrowse",
    "version": 6,
    "params": {
        "query": "deck:current"
    }
}
```
Response
```
{
  "result": null,
  "error": "'MockModel' object has no attribute 'cards'"
}
```

Notably, the gui browser still changes to display "deck:current", but no card ids are returned, just the error. My guess is that this has to do with the browser changes in Anki 2.1.45, since "MockModel" was something introduced into Anki in that release. https://github.com/ankitects/anki/commit/e24f562f8c1f4235225f552cd8d531419303babd

Since my goal is to just change what the gui browser displays, I can probably get away with ignoring the error for now, but I thought I'd bring it up in case it's a problem for anyone else.

ripose-jp:
Anki deprecated access to the model from the API and the new MockModel
contained no attribute 'cards' leading to an error message being thrown
instead of card ids being returned. This change uses findCards to find
card ids instead of the model so no error is thrown while keeping
guiBrowse functionally equivalent to the previous version.
Fixes #277.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -1348,7 +1348,7 @@ class AnkiConnect:
             else:
                 browser.onSearchActivated()
 
-        return list(map(int, browser.model.cards))
+        return self.findCards(query)
 
 
     @util.api()


FooSoft:
Thanks 👍

PR #292, opened 2021-11-20
Add guiSelectedNotes action

penginsan:
This pull request introduces an API action for fetching the notes that are currently selected in the browser.

If the browser is open it returns the identifiers of the selected cards, otherwise it returns an empty list.

My example use case is automating edits to cards. Instead of copying and pasting a field from the card I want the script to detect the card that I'm looking at in the browser.

I have tested the change manually.

Thank you for developing such a useful addon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1018,6 +1018,27 @@ corresponding to when the API was available for use.
     }
     ```
 
+*   **guiSelectedNotes**
+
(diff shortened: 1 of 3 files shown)


penginsan:
Hey @FooSoft any chance you can take a look at this? Should be a quick review.

FooSoft:
Looks good, sorry for the delay!

Issue #301, opened 2022-02-15
Options to handle preflight requests with the Access-Control-Allow-Private-Network header


killergerbah:
According to https://developer.chrome.com/blog/private-network-access-preflight/ Chrome will start requiring servers on private networks to explicitly grant access to public websites through a new header.

> Chrome will start sending a CORS preflight request ahead of any [private network request](https://developer.chrome.com/blog/private-network-access-preflight/#what-is-private-network-access) for a subresource, which asks for explicit permission from the target server. This preflight request will carry a new header, Access-Control-Request-Private-Network: true, and the response to it must carry a corresponding header, Access-Control-Allow-Private-Network: true.

This is supposed to be enforced on version 101, but I am finding that it is already being enforced in some cases for me and I am on version 98. For example, this is now partially breaking AnkiConnect integration for my software [asbplayer](https://github.com/killergerbah/asbplayer). This also breaks the expected behavior of the `requestPermission` API.

PR #302, opened 2022-02-16
Explicitly allow requests from public websites via new header

killergerbah:
- Chrome now enforces that servers on private networks explicitly
      grant access to public websites using a new header
      "Access-Control-Allow-Private-Network" that should be sent in
      responses to preflight OPTIONS requests.
    - This change implements special handling for OPTIONS requests by
      sending all the existing CORS headers along with the new
      Access-Control-Allow-Private-Network header if private network
      access is being requested.
    - See https://developer.chrome.com/blog/private-network-access-preflight/
      for more info.

This PR would fix https://github.com/FooSoft/anki-connect/issues/301

Unfortunately, `requestPermission`  will continue to not work for browsers that send the `OPTIONS` request, as the browser will now fail such requests due to the CORS check resulting from the `OPTIONS` response. A real fix for `requestPermission` could involve bigger change that puts the `requestPermission` action on a different endpoint, so that the `OPTIONS` request for that endpoint could serve a separate origin whitelist.

### DIFF
--- a/plugin/web.py
+++ b/plugin/web.py
@@ -24,7 +24,8 @@ from . import util
 #
 
 class WebRequest:
-    def __init__(self, headers, body):
+    def __init__(self, method, headers, body):
+        self.method = method
         self.headers = headers
         self.body = body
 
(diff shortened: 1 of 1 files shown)


FooSoft:
Great work, thanks!

Issue #286, opened 2021-08-31
 def deleteDecks(self, decks, cardsToo=False):


sosie-js:
Konnitchiwa,

When using Ankisync2 deleteDecks sample from the readme or from the site

result = invoke("deleteDecks",  decks=decks , cardsToo=cardsToo)

it is impossible to delete any deck., it raises Exception(response["error"])
Exception: rem() takes 2 positional arguments but 3 were given

Problem is, I guess it does not match the definition of api function declared __ini__.py
 def deleteDecks(self, decks, cardsToo=False):

as cardsToo initialisation seems to interfers, maybe the same issue with [272](https://github.com/FooSoft/anki-connect/issues/272)

thiswillbeyourgithub:
For reference, here's the code to the new  `remove` function:
https://github.com/ankitects/anki/blob/1cdf0b1e9d86d3ac453118dfb59aa55b90630ece/pylib/anki/decks.py#L147

Bar0-dev:
Calling deleteDecks with javascript as it is on the example with cardssToo: true still doesnt remove the deck. Im going to try to fix this issue soon since I need this fucntion to be working for my project.

oakkitten:
I've been working on fixing the tests, for now i changed this

https://github.com/FooSoft/anki-connect/blob/54a7105bf9da31dabe884a96464206a0fc87c58d/plugin/__init__.py#L532-L541

to say

```python
    @util.api()
    def deleteDecks(self, decks, cardsToo=False):
        if not cardsToo:
            # since f592672fa952260655881a75a2e3c921b2e23857 (2.1.28)
            # (see anki$ git log "-Gassert cardsToo")
            # you can't delete decks without deleting cards as well.
            # however, since 62c23c6816adf912776b9378c008a52bb50b2e8d (2.1.45)
            # passing cardsToo to `rem` (long deprecated) won't raise an error!
            # this is dangerous, so let's raise our own exception
            if self._anki21_version >= 28:
                raise Exception("Since Anki 2.1.28 it's not possible "
                                "to delete decks without deleting cards as well")
        try:
            self.startEditing()
            decks = filter(lambda d: d in self.deckNames(), decks)
            for deck in decks:
                did = self.decks().id(deck)
                self.decks().rem(did, cardsToo=cardsToo)
        finally:
            self.stopEditing()
```

PR #306, opened 2022-03-08
Add Edit dialog

oakkitten:
As discussed in https://github.com/FooSoft/yomichan/issues/2070, this dialog can be an alternative to opening Browser after adding a card in Yomichan. Some parts of the browser, such as the sidebar and the note list, take a lot of space and cannot be collapsed fully. This dialog looks like Edit Current dialog, except it:

- has a Preview button to preview the cards for this note in its regular place
- has a Browse button to open the Browser with them
- has Previous/Back buttons to navigate the history of the dialog
- has no Close button bar

image

This is a work in progress. At the time being the code works fine on Anki ⁨2.1.49 (dc80804a) but has not been tested on other Anki versions. ⁩Remains to be done:

- [x] Add the API
- ~~[ ] Make code compatible with earlier Anki versions~~
- [x] See whether the upcoming Anki version breaks it
- [x] Add tests
- [x] Update documentation

I made some decisions that I'm not completely confident in. 

- I used snake case not camel case. This seems to be what Anki is moving to, and the file is isolated.
- I chose the hotkey `Ctrl-F` for the Browser, and `Alt-Left` and `Alt-Right` for the Previous/Next buttons. `Ctrl-F` is easy to press and that's one of the hotkeys that focus the input field in the Browser. `Alt-Left` and `Alt-Right` are what browsers normally use so I hope these are a good fit. 
- I choose `"foosoft.ankiconnect."` as a prefix for various identifiers, something else might be more appropriate?
- Previous/Next buttons navigate between whatever has been opened in the dialog. Alternatively, it could navigate the cards in the order added.
- If the dialog history is empty and you delete the note in the browser, the Browser fails to reflect the change, keeping the Editor open. It leads to errors if you try to edit the note since it does not exist; however this appears harmless, so I don't suppose this requires much attention? This appears to be a bug in Anki, since the Edit Current dialog is also affected. Not sure what causes it.

Also, what's the minimal version of Anki that the code is supposed to support? This should work on Anki 2.1.45+ without changes. Making it work with earlier versions would lead to significant ugliness.


oakkitten:
I looked into testing a bit. It seems that the best choice would be `pytest` and `pytest-anki`, although it's at the time still in beta. Would it be feasible to move the tests to pytest, and perhaps throw in `poetry` and perhaps CI?

Not sure how to run existing tests, they seem to be failing due to various errors, such as
> Exception: rem() takes 2 positional arguments but 3 were given

or
> Exception: model was not found: Basic

FooSoft:
> I used snake case not camel case. This seems to be what Anki is moving to, and the file is isolated.

I think that's fine. I actually prefer snake case myself (as that follows the python coding standard), but the code all used to be camel case, which I have been matching. Agreed that it just makes sense to use snake case going forward.

> I chose the hotkey Ctrl-F for the Browser, and Alt-Left and Alt-Right for the Previous/Next buttons. Ctrl-F is easy to press and that's one of the hotkeys that focus the input field in the Browser. Alt-Left and Alt-Right are what browsers normally use so I hope these are a good fit.

Sounds good to me.

> I choose "foosoft.ankiconnect." as a prefix for various identifiers, something else might be more appropriate?

I think it's great that there is any sort of identifier at all :) That should work just fine.

> Previous/Next buttons navigate between whatever has been opened in the dialog. Alternatively, it could navigate the cards in the order added.

Fine with either approach.

> If the dialog history is empty and you delete the note in the browser, the Browser fails to reflect the change, keeping the Editor open. It leads to errors if you try to edit the note since it does not exist; however this appears harmless, so I don't suppose this requires much attention? This appears to be a bug in Anki, since the Edit Current dialog is also affected. Not sure what causes it.

Yeah I wouldn't worry about it. It may be fixed in future releases of AnkiConnect for all we know.

oakkitten:
Aha, great! As for testing;

I've been experimenting with testing a bit and if that's fine with you I can try to fix current tests. Most now fail due to changes in Anki. I want to:

* Set minimum supported Anki version to 2.1.45. It was released on 30 Jul 2021 and hopefully most people are using that or a later version;
* Remove some of the API methods that are broken, since these are likely not getting used. (With some exceptions, e.g. #286)
* Move all current tests to `pytest` and test against headless Anki
* Run the tests with `tox` against all... major patches? since 2.1.45
* Integrate with github actions

oakkitten:
Anyway, here's my attempt, with tests running on GitHub Actions [passing](https://github.com/oakkitten/anki-connect/runs/5761539864). I hope the commit history is ok. This PR mostly ready, but I'll keep it as a draft for a few days in case I remember anything.

I removed one API method and fixed a few and converted all tests to pytest and made them run with tox against Anki versions 2.1.45 to 2.1.49. The tests might be slightly flaky. I fixed all of the flakiness I could find but I'm not sure what can be done about segfaults. If the tests prove too flaky, we can try run them forked with `tox -- --forked`. Either way, you should be able to rerun the workflow by pressing a button. Currently the tests run (mostly) in a single process, with the profile being torn down after each test. Actions run so much faster than my computer so if there's any issues due to computer speed I won't be able to deal with those. See `tox.ini` regarding how to run tests.

As for the `Edit` dialog, I refactored it slightly and added a `guiEditNote` method that takes an integer `note`. I'm not sure if the method should attempt to coerce `note` to `int` if it is a `str`; for now I let it fail.

Note that I added GitHub actions for PRs as well. This should be safe as  [due to the dangers inherent to automatic processing of PRs, GitHub’s standard `pull_request` workflow trigger by default prevents write permissions and secrets access to the target repository](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). Just something to keep in mind.

FooSoft:
This is all looking good so far :+1: 
Set the PR is ready when you feel it is in a good spot for merge.

oakkitten:
Thanks! Of course right after pushing I discovered a major problem, that should be fixed now. A few issues should be adressed before this is ready to merge though.

* There's one [skipped test](https://github.com/FooSoft/anki-connect/blob/f72b3bf0a9e585d576ec7803b87260fd437b41c0/tests/test_graphical.py#L86-L111). I want to remove it and disallow calling `guiAddCards` with the argument `closeAfterAdding=True`. While the code does not outright break, the resulting dialog does not do what it is supposed to do anyway. 

* The tests run against Anki 2.1.45-2.1.49 but the plugin might be broken in earlier Anki versions. Is it ok to drop support for Anki < 2.1.45? I think you have multiple versions of an addon (“branches”) on AnkiWeb with different versions supported. I guess targetting 2.1.45+ only will make it easier to support 2.1.50 as well.

Oh and I guess the documentation needs to get updated.

P.S. I set the tests to run with `--forked` and they seem to be very stable in the CI

FooSoft:
I wouldn't worry too much about older versions; they can always use older releases if necessary (or just upgrade to a newer version of AnkiConnect).

oakkitten:
Okay I think this is nearly there (again). I also:

* Made `guiAddCards` ignore the option `closeAfterAdding` I mentioned above;
* Made browser show all the history of the Edit dialog, not just the current note. The table of cards/notes is shown with the most recently opened note on top. The currently opened note, or all cards corresponding to it, are selected. Also, the sorting column indicator is removed to signal that no browser sorting has been performed, but if the user does click on any columns, the sorting works as expected. I guess this worked out pretty awesomely if I do say so myself, and the code isn't even too hacky;
* Figured out a major source of test flakiness and removed it. This actually like a bug in a library used by Anki and might be leading to other issues with Anki, I'll report it there once I get the reply from the library authors.

I'll give it a couple of days and then mark as ready. A few days ago Anki 2.1.50 was released and I want to fix the test to also pass for Anki 2.1.50. Shouldn't be too hard, although since they changed all things JavaScript, adding buttons will require some work.

FooSoft:
Awesome work! I'll put this on AnkiWeb this week.

oakkitten:
Thanks! Although it might have been better not to squash the commits. I specifically tried making the commits in such a way that they are useful for anyone who later looks at git history.

Anyway, I think I will have a few more fixes for Anki 2.1.50 shortly 😛

PR #312, opened 2022-04-13
requestPermission: improve UX & docs

introt:
- Edited popup wording, turned the "Ignore" button into a checkbox (see screenshot below)
- Updated documentation
   - document `ignoreOriginList`
   - described the possible results

image

For reference, see: #254, #255

### DIFF
--- a/README.md
+++ b/README.md
@@ -1503,11 +1503,18 @@ corresponding to when the API was available for use.
 
 *   **requestPermission**
 
-    Request permission to use the API exposed by this plugin. Only request coming from origin listed in the
-    `webCorsOriginList` option are allowed to use the Api. Calling this method will display a popup asking the user
-    if he want to allow your origin to use the Api. This is the only method that can be called even if the origin of
-    the request isn't in the `webCorsOriginList` list. It also doesn't require the api key. Calling this method will
-    not display the popup if the origin is already trusted.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1503,11 +1503,18 @@ corresponding to when the API was available for use.
...
+    only one that accepts requests from any origin; the other methods only accept requests from trusted origins,
+    which are listed under `webCorsOriginList` in the add-on config. `localhost` is trusted by default.
+
+    Calling this method from an untrusted origin will display a popup in Anki asking the user whether they want to
+    allow your origin to use the API; calls from trusted origins will return the result without displaying the popup.
+    When denying permission, the user may also choose to ignore further permission requests from that origin. These
+    origins end up in the `ignoreOriginList`, editable via the add-on config.
+
+    The result always contains the `permission` field, which in turn contains either the string `granted` or `denied`,
+    corresponding to whether your origin is trusted. If your origin is trusted, the fields `requireApiKey` (`true` if
```
introt, line 1516: If we were to return the API key along with the permission, the field could be renamed to _require**d**ApiKey_, which would then contain either the key, or `null` if one isn't required. Given the required API version bump, It'd also be a good opportunity to change `permission` (string) to `permissionGranted` (boolean).

FooSoft:
Thanks!

PR #318, opened 2022-05-03
Add getDeckStats action

ZhengPeiRu21:
This PR adds an action that allows the API to get statistics for given decks. This is useful for consumers that would like to know the number of cards already in a deck or the daily new cards or reviews remaining.


FooSoft:
Awesome, thanks!

PR #319, opened 2022-05-12
Produce a better error in case of malformed JSON

oakkitten:
This solves the same issue that #317 solves, albeit slightly differently. I was hoping that the author would reopen the issue... There's a few differences;

* If origin is not allowed, 403 is returned in case of malformed JSON, and
* In case of JSON error, error string is returned to user.

I also noticed this bit: `params['params'] = params.get('params', {})`. It's a tad dangerous since this part of code can be run when not allowed, and `params` can be present but not be a dictionary. In this case the plugin would “crash”, and while this only amounts to showing a window with an error, it can be theoretically triggered by a malicious website, so I opted for verifying request schema. (Anki comes with `jsonschema` which is used to verify addon configurations.) This is not ideal, but I wanted to touch as little code as possible.

I added a few tests that test behavior around the change.

@nvlled Could you please look at this and tell if I did something stupid? 😬

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -43,6 +43,7 @@ from anki.notes import Note
 from anki.errors import NotFoundError
 from aqt.qt import Qt, QTimer, QMessageBox, QCheckBox
 
+from .web import format_exception_reply, format_success_reply
 from .edit import Edit
 from . import web, util
 
@@ -99,7 +100,6 @@ class AnkiConnect:
         version = request.get('version', 4)
         params = request.get('params', {})
         key = request.get('key')
-        reply = {'result': None, 'error': None}
 
         try:
             if key != util.setting('apiKey') and name != 'requestPermission':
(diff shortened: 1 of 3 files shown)


FooSoft:
Looks great to me :)

oakkitten:
I really wish @nvlled said something, it now feels like I stole their PR :<

nvlled:
Sorry, I couldn't find the time to give a thorough review. But I do one nit: explicity initialize params outside of the try block. Something like:
```python
params = {}
try:
    params = json.loads(req.body.decode('utf-8'))
```

The bug I encountered was specifically cause by an uninitialized params variable. It would be nice to prevent this from happening when changes are made in the future.

oakkitten:
Ah, that should be solved in this PR! In the `except` part of the of the block, if continuing, the `params` variable is set. I suppose one could also set it before like in your example. It's just that I use PyCharm and I'm accustomed to it warning me if I forget to set a variable in one of the execution branches.

https://github.com/FooSoft/anki-connect/blob/9c3310a8b6202f93ed42ec13461566659b1e2036/plugin/web.py#L179-L192

I'd still appreciate the review! Anyway, I guess I better make this not a draft since, well, the button says “Ready for review” :p

FooSoft:
Looks good to me!

oakkitten:
P.S. the tests are failing now due to protobuf update (packages anki/aqt don't pin dependencies), I'll have a fix shortly

FooSoft:
Thanks!

PR #322, opened 2022-05-16
Fix getDeckStats test

ZhengPeiRu21:
Sorry that I did not see this earlier! When I first implemented getDeckStats, it supported only one deck per call. When updated to support multiple decks I did not properly update the unit test. This should resolve the CI build error due to the test failing after https://github.com/FooSoft/anki-connect/pull/318. I apologize for any trouble!


oakkitten:
It seems that the test only fails on Anki 2.1.46 and below;  does `total_in_deck` exist on Anki 2.1.46?

Also, you can run tests yourself. See the comment in `tox.ini` for instructions

oakkitten:
Oh, and this change has way too many lines changed. Perhaps an issue with line endings?

ZhengPeiRu21:
It does seem to being a line ending problem. I will try to fix it.

oakkitten:
I'm no authority here, but if you asked me, there are four ways, in order of least preference:

  1. Remove it;
  2. Don't include it in the results if running on Anki that doesn't support it, or return some useless value, and add a note in the comment;
  3. If it is computable, compute it where necessary and include in the results. We've got `anki_version` that you can check against;
  4. Figure out if maybe it's worth bumping out the minimum supported Anki version.

Anki mixes “stable” versions (e.g. 2.1.49) and not-quite-stable (e.g. 2.1.50). I'm not sure if 2.1.45 and 2.1.46 are stable, or in other words, that many users use these versions. If we can figure which versions are not stable, we can drop support of them. That would be the best solution I guess? (Users who do use versions like 2.1.45 will still be able to use Anki-Connect, just without the new updates.)

ZhengPeiRu21:
I have added a version check and only add the `total_in_deck` value in supported versions. It seems the line endings are still not correct; I will fix this next. Thank you for your helping!

ZhengPeiRu21:
I have fixed the line endings. Thank you for the patience!

oakkitten:
Nice! Perhaps also update the readme to say that `total_in_deck` will only be returned on api 2.1.47+. 

It would also be nice to test some numbers maybe? The tests run in isolation so you can count on them being zeros. Or you can test using [the setup fixture](https://github.com/FooSoft/anki-connect/blob/1d7bf6969a23921336d197cc9d69fd35c37ae700/tests/conftest.py#L285-L299), it has a few cards.

(P.S. This is a matter of personal taste; I prefer to say `if anki_version >= (2, 1, 47)` not `if anki_version > (2, 1, 46)`, I just it's nice to refer to the “the first version that supports“ than to “the last version that doesn't support”.)

PR #342, opened 2022-07-18
Feat: add `interpret` action

skeledrew:
This PR adds the `interpret` action, which allows code to be run in the context of the add-on. This allows consumers to perform arbitrary actions, without the need to define specific API points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1753,6 +1753,29 @@ corresponding to when the API was available for use.
     }
     ```
 
+*   **interpret**
+
+    Executes arbitrary code in the plugin's namespace and returns the result or `null` if given a statement. If the result cannot be JSON-encoded, it is stringified.
+
+    *Sample request*:
+    ```json
+    {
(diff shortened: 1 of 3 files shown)


FooSoft:
As is, this is a big security risk. Someone can just execute arbitrary Python code as system user by sending a request to Anki that uses this proposed action.

skeledrew:
That is true. Any ideas how it can be made safer? My use case is I'm working on a package that creates a bridge web server for the Anki desktop app interface, usable by regular browsers particularly on mobile.

FooSoft:
Probably a step to explicitly allow this action in the config would be required. It would also have to be made very obvious to the user that their computer can be entirely hijacked by doing so. There is no way this can be enabled by default; we have to make sure to educate non-technical users of Anki as to what this would allow.

FooSoft:
Closing due to lack of activity, feel free to reopen if this changes.

PR #343, opened 2022-07-25
Add PowerShell example to Readme.

StevenTCramer:
Now that PowerShell is cross platform it is my go to shell.  Thought I would add an example to the readme.


FooSoft:
Thanks!

PR #346, opened 2022-08-17
Added model field name actions

Aquafina-water-bottle:
## Use case
I've been working on a script to update a model in place, and I occasionally modify the field names (positioning, etc.) of the model. It would be convenient to use AnkiConnect to do just that.

## Additional notes
- ~~I'm currently working on fleshing out the documentation, hence why this is a draft request.~~ edit: documentation added.
- Seeing that the `multi` action exists, I figured that each action only affecting one field is sufficient.
- I'm personally conflicted on the exact naming of the actions. Perhaps something along the lines of `renameFieldName` or `renameModelFieldName` would be better. Do you have any ideas?
- I was wondering if certain errors should be silent. Currently, the `modelFieldAdd` action does not show an error if the field already exists, and continues to reposition the field. However, the `modelFieldRemove` (and the other 2 actions) will show an error if the field does not already exist. Is this behavior good?

Thanks!


FooSoft:
Hey @Aquafina-water-bottle I think this PR looks good. The API surface (for better or worse) has grown organically, and some things are more consistent than others. I think naming and error handling is fine the way you have it now. Thanks for the contribution!

PR #350, opened 2022-09-05
Added getReviewsOfCards

Aquafina-water-bottle:
This introduces an API action to get review data of the specified cards, compared to `cardReviews` which returns all the reviews after a given time.

## Use case
Although I'm not completely sure what other practical applications this can have, for me in particular, I'm interested in getting the earliest review time of a card to sort by for a small application in my Anki template. I figured that instead of making a specific API call to get the earliest / latest review time of a card, a more general API action like this should suffice.

## Additional Notes
Again with naming, given that there is a `cardReviews` call that already exists, I'm hoping this name is good. If you have any suggestions for potentially better names that is sufficiently different from `cardReviews`, please let me know!

Thanks!

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -1374,6 +1374,13 @@ class AnkiConnect:
 
 
     @util.api()
+    def getReviewsOfCard(self, card):
+        return self.database().all(
+            'select id, cid, usn, ease, ivl, lastIvl, factor, time, type from revlog where cid = ?', card
+        )
+
+
+    @util.api()
     def reloadCollection(self):
         self.collection().reset()
 
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -1375,13 +1375,15 @@ class AnkiConnect:
 
     @util.api()
     def getReviewsOfCards(self, cards):
-        return {
-            card: self.database().all(
-                "select id, cid, usn, ease, ivl, lastIvl, factor, time, type from revlog where cid = ?",
-                card,
-            )
-            for card in cards
-        }
+        COLUMNS = ['id', 'usn', 'ease', 'ivl', 'lastIvl', 'factor', 'time', 'type']
+        QUERY = 'select {} from revlog where cid = ?'.format(', '.join(COLUMNS))
+
+        result = {}
+        for card in cards:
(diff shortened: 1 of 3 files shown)


plugin/__init__.py
```diff
@@ -1373,6 +1373,13 @@ def cardReviews(self, deck, startID):
         )
 
 
+    @util.api()
+    def getReviewsOfCard(self, card):
+        return self.database().all(
+            'select id, cid, usn, ease, ivl, lastIvl, factor, time, type from revlog where cid = ?', card
```
FooSoft, line 1379: Can you change this to be a dictionary instead of a simple list? Using arbitrary indices is error prone.

FooSoft (changes requested):

plugin/__init__.py
Aquafina-water-bottle: Sounds good! As an additional note, I changed the API call to accept multiple cards in its argument, since the dictionary format is a lot safer to use, and there seems to be enough of a precedent in the API to read values from multiple cards at once.

FooSoft:
I see you made some changes, here but the inner lists are still not dictionaries.

```

            "1653613948202": [
                [1654102387663, 1653613948202, 1780, 3,   8,  3, 2500, 25796, 1],
                [1654798974478, 1653613948202, 1861, 3,  20,  8, 2500, 18134, 1],
                [1656556319328, 1653613948202, 2075, 3,  53, 20, 2500, 20530, 1],
                [1661107990069, 1653613948202, 2478, 3, 131, 53, 2500, 24247, 1]
            ]
```

Aquafina-water-bottle:
Sorry for the delay! I misunderstood the request from earlier, but now the result is the following:
```json
{
    "1653613948202": [
        {
            "id": 1653772912146,
            "usn": 1750,
            "ease": 1,
            "ivl": -20,
            "lastIvl": -20,
            "factor": 0,
            "time": 38192,
            "type": 0
        },
        {
            "id": 1653772965429,
            "usn": 1750,
            "ease": 3,
            "ivl": -45,
            "lastIvl": -20,
            "factor": 0,
            "time": 15337,
            "type": 0
        }
    ]
}
```
Additionally, I removed the `cid` section within the dictionaries.

Issue #352, opened 2022-09-11
Duplicate check fails when media added to the Front Field


SimbaNinja917:
Hello,

When I did the API POST request below to add an _Audio+Text_ note to my deck, several times, the duplicate check fails and note is added multiple times. This happens only when media (audio in my case) is in the Front field. 

I delete the notes from my deck and perform the following test: 
I delete the "audio" JSON object from the JSON request body from the _Audio+Text_ note. As a consequence, it becomes a _text-only_ note. I send an API POST multiple times, and duplicate check works successfully, and _no_ duplicate _text-only_ note is added. 

I delete the notes from my deck and repeat a different test: 
When I try again _Audio+Text_ note _after_ API POSTing a _text-only_ note, duplicate check prevents the _Audio+Text_ note from being added.  

## Investigating what's happening 
I did some checks in Anki's SQLite database, and tracked the checksums in the tables. I also reverse engineered the `fieldChecksum` method after looking at `__init__.py` and `anki/utils.py`. 

## Here are the comparisons of checksums using POST API requests
checksum value in SQLite database for Front _Audio+Text_ `front content[sound:yomichan.mp3]` is: `3144727897`
checksum in SQLite database for Front _text-only_ `front content` is: `3787401650`
checksum in SQLite database for Front _text-only_ `front content[sound:yomichan.mp3]` is: `3144727897`

adding _text-only_ note when _text-only_ note is in SQLite database: 3787401650 (request) vs 3787401650 (in database) duplicate check succeeds (works as it should be) ✅. 

adding _sound+text_ when _sound+text_ is in SQLite database: "supposed to be" 3144727897 (request) vs 3144727897 (in database): duplicate check fails (false negative) ⚠️. **This means that checksums aren't matching.**  

adding _sound+text_ when _text-only_ is in SQLite database: "supposed to be" 3144727897 (request) vs 3787401650 (in database): wrongly detects duplication (false positive) ⚠️. **This means that checksums are matching.** 

adding _text-only_ when _sound+text_ is in SQLite database: 3787401650 (request) vs 3144727897 (in database): duplicate check as expected (no duplicate detected) ✅.

adding a _text-only_ note with the front field value `front content[sound:yomichan.mp3]` when _sound+text_ is in SQLite database: 3144727897 (request) vs 3144727897 (in database): duplicate check positive ✅. **This means that checksums are matching**. 

[...]

PR #353, opened 2022-09-12
Fix issue #352

snoyes:
Adding media might alter the content of the first field. The duplicate check had been run before adding media; now the media is added first.


FooSoft:
Thanks!

PR #363, opened 2022-11-26
Added modelFieldDescriptions

introt:
Similar to modelFieldNames, returns a list of note field decriptions (the text seen in the gui editor when the field is empty).

Closes #362 (in favor of a new issue targetting the creation of field descriptions via anki-connect).

Issue #362, opened 2022-11-26
Feature request: Note field descriptions


introt:
Note field descriptions are the "Text to show inside the field when it's empty". They're useful for in-app note template documentation which in turn reduces the cognitive barrier to systematic note creation (eg. "does this factoid belong in the note field named ["etiology" or "pathology"](https://medicalsciences.stackexchange.com/questions/1865/whats-the-difference-between-etiology-pathogenesis-pathology-pathophysiology)?).

Relevant `anki` source files:

```python
qt/aqt/fields.py:        f.fieldDescription.setText(fld.get("description", ""))
qt/aqt/fields.py:        desc = f.fieldDescription.text()
qt/aqt/fields.py:        if fld.get("description", "") != desc:
qt/aqt/fields.py:            fld["description"] = desc
qt/aqt/forms/fields.ui:      <widget class="QLineEdit" name="fieldDescription">
qt/aqt/forms/fields.ui:        <string>fields_description_placeholder</string>
qt/aqt/forms/fields.ui:  <tabstop>fieldDescription</tabstop>
rslib/src/notetype/fields.rs:                description: "".into(),
ftl/core/fields.ftl:fields-description = Description
ftl/core/fields.ftl:fields-description-placeholder = Text to show inside the field when it's empty
ts/editor/NoteEditor.svelte:    let fieldDescriptions: string[] = [];
ts/editor/NoteEditor.svelte:        fieldDescriptions = fs;
ts/editor/NoteEditor.svelte:        description: fieldDescriptions[index],
ts/editor/NoteEditor.svelte:                        --description-content={`"${field.description}"`}
```
(introduced in ankitects/anki#1476)

Adding support requires certain considerations, as currently fields are mainly treated as ordered lists (only).

My suggestion is to start with `modelFieldDescriptions` similar to `modelFieldNames` on the querying side (implementation [here](https://github.com/FooSoft/anki-connect/issues/362#issuecomment-1328037776)). Besides that, adding `inOrderFieldDescriptions` to `createModel` would be one way to proceed in the other direction, followed by `modelFieldDescriptionRename` akin to `modelFieldRename`.

introt:
Related: `createModel` returns the editing font, direction (rtl) and stickyness, but setting nor querying them is not supported:

```json
"flds":[
            {
                "name":"Field1",
                "ord":0,
                "sticky":false,
                "rtl":false,
                "font":"Arial",
                "size":20,
                "media":[

                ]
            },
            {
[...]

PR #367, opened 2022-12-24
Added various calls for model fields (getting & setting font, setting description)

Aquafina-water-bottle:
This PR picks up the work from #362 and #363, and adds the following calls:
- `modelFieldFonts`
- `modelFieldSetDescription`
- `modelFieldSetFontSize`
- `modelFieldSetFont`

Additionally, some helper methods were added to prevent repeated code. So far, only the model field related calls have been changed to use the helper methods.

TODO:
- [x] Add tests
- [x] Verify tests
- [x] Test what happens if `modelFieldSetDescription` is used on older versions of Anki
- [x] Remove commented code
- [x] Add documentation to README
- [x] Final formatting check

Note:
- Currently, `modelFieldFonts` returns a json of the following format:

      [
        {
          "fieldName1": {
            "font": "fontName",
            "size": fontSize
          }
        },
        ...
      ]

  However, this deviates from the related `modelFieldDescriptions` call, that simply returns a list of descriptions. I was wondering if this is still fine?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2246,6 +2246,30 @@ corresponding to when the API was available for use.
     }
     ```
 
+*   **modelFieldRemove**
+
+    Deletes a field within a given model.
+
+    *Sample Request*:
+    ```json
+    {
+        "action": "modelFieldRemove",
+        "version": 6,
+        "params": {
+            "modelName": "Basic",
+            "fieldName": "Front"
+        }
+    }
+    ```
+
+    *Sample result*:
+    ```json
+    {
+        "result": null,
+        "error": null
+    }
+    ```
+
 *   **modelFieldSetFont**
 
     Sets the font for a field within a given model.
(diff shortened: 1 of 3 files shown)


FooSoft:
@Aquafina-water-bottle I think that's perfect fine, and is probably in fact a better way of doing things.

Aquafina-water-bottle:
Some additional notes:
- I changed the format of `modelFieldFonts` to return a dictionary format so fields can be accessed easier i.e.:
  ```json
  {
      "field1": {
          "font": "Arial",
          "size": 20,
      },
      "field2": {
          "font": "Arial",
          "size": 16,
      },
  }
  ```
- In `modelFieldSetDescription`, it doesn't set the description if the key doesn't exist. This appears to be safe, and is properly tested in the tests. Additionally, when I made a new profile in Anki 2.1.49 and 2.1.54, the call still works as expected and properly sets the description.

Aquafina-water-bottle:
This should be ready for review now! The tests that fail seem to be failing at random, so I believe it should be safe to merge if everything looks good.

PR #369, opened 2022-12-29
Add getNoteTags, updateNoteTags and updateNote

introt:
Before writing README docs, are these ok? This updateNoteTags implementation is quite wasteful, but doing it like replaceTags kept not deleting tags.

getNoteTags returns [str ...] of tags for a given note (nid: int)

updateNoteTags removes the existing tags and sets new tags

updateNote combines the old updateNoteFields and the new updateNoteTags

I'd love to have these included, as implementing these client side requires a bunch of API calls. Will redo commit msg when adding docs.

### DIFF
--- a/plugin/__init__.py
+++ b/plugin/__init__.py
@@ -793,6 +793,30 @@ class AnkiConnect:
 
         self.collection().autosave()
         self.stopEditing()
+    
+    
+    @util.api()
+    def updateNote(self, note):
+        self.updateNoteFields(note)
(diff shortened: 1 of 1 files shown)


FooSoft:
I don't see any problems with this. For better or worse, the AnkiConnect API has evolved organically based on specific requirements, so there are some holes like this. Let me know when PR is ready and I'll merge it in.

introt:
Closed in favor of #370

PR #373, opened 2023-01-14
Reformat README to be more readable

Aquafina-water-bottle:
This PR reformats the README to simply do the following:
- Hide the sample requests/results with a details tag
- Move the action names from a list bullet point to a h4 title
- And some very minor stylistic changes, such as adding a line between the action categories to better separate them.

To see how the changes look in action, click [here](https://github.com/Aquafina-water-bottle/anki-connect/tree/readme_reformat#supported-actions).

I personally find the documentation in the README difficult to read due to the amount of vertical scrolling I have to do, and it seems like it will only get worse as the API expands. Hiding the sample requests and results prevents a lot of extra scrolling, and allows the user to read the actual documentation behind the API easier. Additionally, moving the action names to a h4 title has the added benefit of being able to permalink to any specific action.

Some notes:
- These changes were inspired by how [Yomichan's template helpers are already documented](https://github.com/FooSoft/yomichan/blob/master/docs/templates.md).
- I don't know if this will look good on the [foosoft.net](https://foosoft.net/projects/anki-connect/) website (especially with the new line divisions between the action categories). It might be a good idea to see how the changes look there beforehand.

As with all stylistic changes, there are bound to be differing opinions, so I would love to hear any feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2369,7 +2369,6 @@ corresponding to when the API was available for use.
             "mod":1551462107,
             "usn":-1,
             "vers":[
-            </details>
 
             ],
             "type":0,
@@ -2427,6 +2426,7 @@ corresponding to when the API was available for use.
         "error":null
     }
     ```
+    </details>
 
 #### `modelTemplates`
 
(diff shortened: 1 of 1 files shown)


FooSoft:
Ooh! I like this. The markup probably won't look this good initially to start foosoft.net, but I have no problems styling it :)

PR #375, opened 2023-02-12
Return deckName in notesInfo.

Akida31:
If I'm not wrong there was previously no way to get the name (or id) of the deck the note belongs to.

I've tested my changes with python3.10. Is there anything else I have (or can) do?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3308,6 +3308,7 @@ corresponding to when the API was available for use.
             {
                 "noteId":1502298033753,
                 "modelName": "Basic",
+                "deckName": "Default",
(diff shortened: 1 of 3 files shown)


Akida31:
Ok I discovered that my approach didn't work in general. It returns the deck associated with the model and not with the note. This became a problem if the model is used in multiple decks.
Instead I found a (probably hacky) way around and maybe it is interesting for others:
You can use 'cards' in 'notesInfo' to get the ids of all associated cards and check then (with 'cardsInfo') that all cards have the same deck and use this

PR #376, opened 2023-02-18
Changed the way to bind to an address other than the default 127.0.0.1

kentonself:
The code does not look at ANKICONNECT_BIND_ADDRESS in the environment. It gets it from the config.json which can be editted from the Tools menu.


FooSoft:
Thanks!

kentonself:
Thanks for merging!

PR #379, opened 2023-03-21
add a new action: getMediaDirPath

tatsumoto-ren:
This PR adds the `getMediaDirPath` action that returns the full path to the user's collection folder. 

Currently Mpvacious has to rely on the `storeMediaFile` action to add new audio and images to the collection, but this [proves to be problematic](https://github.com/Ajatt-Tools/mpvacious/issues/94). When called with base64-encoded data, it may fail if the file is too big, and when called with a path to a temporary file, it may fail because Anki doesn't have access to the folder where the file is stored.

With this function we would be able to send files directly to the `collection.media` folder without encoding them as base64 or using temporary files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1759,6 +1759,32 @@ corresponding to when the API was available for use.
     ```
     </details>
 
+#### `getMediaDirPath`
+
+*   Gets the full path to the `collection.media` folder of the currently opened profile.
+
+    <details>
+    <summary><i>Sample request:</i></summary>
+
+    ```json
+    {
+        "action": "getMediaDirPath",
+        "version": 6
+    }
+    ```
+    </details>
+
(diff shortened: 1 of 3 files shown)


FooSoft:
Looks good, could you just add a simple test case for calling this API and making sure it doesn't fail?

tatsumoto-ren:
I can try. This repo doesn't include any docs on how to create and run tests, I'll try inferring that from the source code. 

Seems like I need a program called `tox`, never heard of it.

FooSoft:
Thanks!

tatsumoto-ren:
The extension on AnkiWeb hasn't been updated yet. It would be awesome if you uploaded a new version 🙏.

Issue #378, opened 2023-03-18
Cannot use api "getReviewsOfCards"


dlangevi:
Wanted to make use of this api call, but it appears it has not been released throught the Anki plugin system (seems that the last update was in September 2022) 

Do you have a timeline of when the latest release found on github will be pushed to anki users?

Aquafina-water-bottle:
FYI, I believe the newest version on AnkiWeb has the action now \o/

PR #380, opened 2023-03-31
Update note

Aquafina-water-bottle:
It's been a while since the author of #370 did anything, and I'm interested in using these actions at some point in the future. This PR is simply #370 without merge conflicts. All credits go to @introt for the new features.

P.S. Echoing #378, is there a release plan for Anki-Connect to AnkiWeb? It's been a little while since the last update to the AnkiWeb version, and I would also like to use some of the recent changes at some point. Thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2533,7 +2533,7 @@ corresponding to when the API was available for use.
 
 *   **updateNoteFields**
 
-    Modify the fields of an exist note. You can also include audio, video, or picture files which will be added to the note with an
+    Modify the fields of an existing note. You can also include audio, video, or picture files which will be added to the note with an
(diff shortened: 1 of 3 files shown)


FooSoft:
Looks good, thanks!

introt:
Thanks @Aquafina-water-bottle !

Aquafina-water-bottle:
@schlauchBot Thanks for the tip! I'm using the master branch version with `link.sh` already, so I do indeed have the changes. The concern is primarily for other people who may want to use programs that require the latest features (I think the guys behind mpvacious wants to use the `getMediaDirPath` call for example). It would be a hassle to get their users to download a custom version of AnkiConnect just to use that.

PR #382, opened 2023-04-14
Added various actions relating to model templates

Aquafina-water-bottle:
This PR adds a few actions to modify the card template list itself, and should be mostly analogous to the modelField actions. These actions should make it possible to edit a model in place with even more precision than before, and is necessary if we want to do something like add another card template to the model automatically. This use case is exactly what I want to do sometime in the future with my note.

Additionally, this PR makes a minor change to the internal `getField` function, to reduce a bit of redundant code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2630,6 +2630,134 @@ corresponding to when the API was available for use.
     ```
     </details>
 
+#### `modelTemplateRename`
+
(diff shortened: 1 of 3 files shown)


FooSoft:
Looks good, thanks!

Issue #386, opened 2023-05-27
`insertReviews()` function


JacobH140:
Can the insertReviews() function before to simulate answering cards programmatically? I’m not able to find much information online regarding it, and I am not confident enough in my knowledge of Anki’s review algorithm to attempt replicating its behavior on all 9 of the required inputs for each review (see the ‘Manual Analysis’ subsection of the docs).

PR #390, opened 2023-06-14
add new action: answerCards

### LINKED ISSUE #148: Feature Request: answerCard action
Currently API exposes operation `guiAnswerCard`. This operation depends on current GUI state.

In my opinion similar operation, but independent to GUI state, is missing.

Example request:

```
{
    "action": "answerCard",
    "version": 6,
    "params": {
        "ease": 1,
        "cardId": 1498938915662
    }
}
```

antistic:
fixes #148

might address #386

thanks for the plugin :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -726,7 +726,7 @@ corresponding to when the API was available for use.
 
 #### `answerCards`
 
-*   Answer cards. Answers are between 1 (Again) and 4 (Easy). Will start the timer immediately before answering. Returns `true` if card exists, `false` otherwise.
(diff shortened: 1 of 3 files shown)


FooSoft:
Thanks for the PR! Could you make it so that instead of having two parallel lists, the API instead takes a single list of objects that have a cardId and answer value? That will be more robust than trying to make sure that the lists are always have the same # of elements and are lining up correctly.

antistic:
Sure, I initially did it that way to match with `setEaseFactors` but you're right a list of objects makes it easier.

antistic:
i went with the last one but called it "ease" since that matches the anki code, though i'm happy to change it to "button" or something else

FooSoft:
Looks good to me!

Issue #389, opened 2023-06-12
Requests on Python + Windows are always around ~2s long


Aquafina-water-bottle:
Hey all,

I believe there's a strange problem where sending Anki-Connect requests on Python+Windows takes about 2 seconds. I tried using `urllib.request` and the `requests` library, but they both get the same result (time of around 2 seconds). I sent the code for `urllib.request` to someone else, and they were able to reproduce the issue. This issue appears to also be present when disabling Windows defender, and downgrading to Python 8 + Anki 2.1.49.

The strangest part is that I cannot reproduce this with any other combination, whether it is JavaScript on Windows, PowerShell on Windows, Python on Linux, curl on Linux, etc.

Example code:
```
import urllib.request
import time

start = time.time()
x = urllib.request.urlopen(urllib.request.Request("http://localhost:8765", {}))
end = time.time()
print(x, end-start)
```

```
import requests
import time

start = time.time()
x = requests.post("http://localhost:8765", "{}")
end = time.time()
print(x, end-start)
```

If anyone happens to know the source of the issue, a workaround to this, or even a simple lead on why this happens in the first place, please let me know!

Aquafina-water-bottle:
Seems like this has to do with how `localhost` is resolved on Windows:
- https://stackoverflow.com/questions/25199405/localhost-vs-127-0-0-1-performance
- https://stackoverflow.com/questions/47766158/why-is-python-requests-to-localhost-slow

The main workaround is to use `127.0.0.1` instead of `localhost`.

PR #391, opened 2023-06-15
Updated Python example code to use 127.0.0.1 instead of localhost

Aquafina-water-bottle:
Apparently, `urllib.request.urlopen` takes a very long time to load when using `localhost` instead of `127.0.0.1` on Windows. This PR simply changes `localhost` to `127.0.0.1` in the sample code.

I initially thought of doing something like:
```
import platform
url = "127.0.0.1" if platform.system() == "Windows" else "localhost"
```
However, given that the Javascript example is using 127.0.0.1 already, I figured it is fine to replace `localhost` entirely.

Related issue & more reading material: #389

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ def request(action, **params):
 
 def invoke(action, **params):
     requestJson = json.dumps(request(action, **params)).encode('utf-8')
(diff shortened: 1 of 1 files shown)


FooSoft:
How bizarre! Anyways, thanks for the fix!

PR #392, opened 2023-06-17
Add a method to bring up Import File dialog

mikkkee:
Hi, 

Thanks for this great plugin! Hope you can have a look at this PR when you have a moment.

Currently AnkiConnect only supports importing of `apkg` file. This PR adds the support to import all file types by adding a GUI action to bring up the "Import... (Ctrl+Shift+I)" dialog.

If a path is provided, it brings up an "Import File" window with provided file. If no path is provided, it brings up select file window followed by "Import File" window. 

To avoid being overlooked by the user, Anki window is brought to front in this action.

See below screenshot for the "Import File" window brought up by the following command in PowerShell. 

```
(Invoke-RestMethod -Uri http://localhost:8765 -Method Post -Body '{"action": "guiImportFile", "version": 6, "params": {"path": "C:/Users/Downloads/anki-import.txt"}}')
```

image

Thanks a lot.

Please kindly let me know if any comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1617,6 +1617,35 @@ corresponding to when the API was available for use.
     ```
     </details>
 
+#### `guiImportFile`
+
+*   Invokes the *Import... (Ctrl+Shift+I)* dialog with an optional file path. Brings up the dialog for user to review the import. Supports all file types that Anki supports. Brings open file dialog if no path is provided. Forward slashes must be used in the path on Windows.
+
+    <details>
+    <summary><i>Sample request:</i></summary>
+
+    ```json
+    {
+        "action": "guiImportFile",
+        "version": 6,
+        "params": {
+            "path": "C:/Users/Desktop/cards.txt"
+        }
+    }
+    ```
+    </details>
+
+    <details>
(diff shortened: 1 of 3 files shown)


FooSoft:
Looks good, could you please add a test that just calls this API?

mikkkee:
> Looks good, could you please add a test that just calls this API?

Hi thanks for your comments but I'm not sure how to run tests for an Anki plugin. 
Does it work if I simply add the following function to `/tests/test_graphical.py` ?

```python
def test_guiImportFile(setup):
    ac.guiImportFile()
```

FooSoft:
Yep, assuming that passes, something like that should be just fine!

mikkkee:
> Yep, assuming that passes, something like that should be just fine!

Thank you! just committed the test

mikkkee:
oops looks it breaks the test

```
 plugin/__init__.py:44: in <module>
    from aqt.import_export.importing import import_file, prompt_for_file_then_import
E   ModuleNotFoundError: No module named 'aqt.import_export'
```

PR #394, opened 2023-06-23
Correct JSON MIME type (Fixes #393)

### LINKED ISSUE #393: Correct JSON MIME type
Correct MIME type for JSON is [`application/json`](https://en.wikipedia.org/wiki/Json), not `text/json`. In aiohttp, a manual override is required for every request, which is inconvenient:

```py
response.json(content_type='text/json')
```

nihil-admirari:


FooSoft:
Thanks!

PR #396, opened 2023-07-06
Fix #392 with version check for Anki and Qt

mikkkee:
Hi, 

Some tests are failing due to #392 , there are two kinds of errors

1. `aqt.import_export` doesn't exist, it turns out this is only added in Anki 2.1.52. See #1743 in https://github.com/ankitects/anki/compare/2.1.51...2.1.52 
2. `Qt.WindowStaysOnTopHint` doesn't exist, it turns out it's of a different name in Qt6. 

This pull request is to handle both and update doc to make it clear `guiImportFile` is only supported >=2.1.52

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1619,7 +1619,7 @@ corresponding to when the API was available for use.
 
 #### `guiImportFile`
 
-*   Invokes the *Import... (Ctrl+Shift+I)* dialog with an optional file path. Brings up the dialog for user to review the import. Supports all file types that Anki supports. Brings open file dialog if no path is provided. Forward slashes must be used in the path on Windows.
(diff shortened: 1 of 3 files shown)


mikkkee:
Oh looks I need to add a button click after bringing up the import window to avoid timeout

mikkkee:
Tests are good now 😄

FooSoft:
Nice, thanks for fixing up the back compat!

mikkkee:
This test fails, but it looks not related to this change.

https://github.com/FooSoft/anki-connect/actions/runs/5495042567/jobs/10014182696

PR #403, opened 2023-09-24
Make it slightly clearer how to enable authentication.

Julian:
The README previously sort of covered this, but in particular it was not obvious to me that the request parameter that corresponded to enabling `apiKey` is called `key`, and not `apiKey`, which ultimately made me have to look at the source code to figure this out (even though the word `key` is mentioned elsewhere on the page, it still wasn't very easy to make the connection).

This section puts all the required information in one spot.


FooSoft:
Thanks!

